PR #2, opened 2021-06-12
Fix missing includes of <limits>

lf-:
Example compile error from gcc-11.1.0 on archlinux:
```
/home/jade/builds/brlcad/src/brlcad/src/libged/brep/brep.cpp: In function ‘int _brep_cmd_plate_mode(void*, int, const char*
*)’:
/home/jade/builds/brlcad/src/brlcad/src/libged/brep/brep.cpp:810:48: error: ‘numeric_limits’ is not a member of ‘std’
  810 |     ss << std::fixed << std::setprecision(std::numeric_limits<double>::max_digits10) << pthicknessmm;
      |                                                ^~~~~~~~~~~~~~
```

### DIFF
--- a/src/libbg/earcut.hpp
+++ b/src/libbg/earcut.hpp
@@ -24,6 +24,7 @@
 #include <algorithm>
 #include <cassert>
 #include <cmath>
+#include <limits>
 #include <memory>
 #include <vector>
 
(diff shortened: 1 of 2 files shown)


starseeker:
Good catch - thanks!

PR #6, opened 2021-11-03
Updated comments to Doxygen format

redcarrera:
Includes minor code format changes

### DIFF
--- a/src/external/Creo/assembly.cpp
+++ b/src/external/Creo/assembly.cpp
@@ -1,4 +1,5 @@
-/*                  A S S E M B L Y . C P P
+/*
+ *                  A S S E M B L Y . C P P
  * BRL-CAD
  *
  * Copyright (c) 2017-2021 United States Government as represented by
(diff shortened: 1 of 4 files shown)


brlcad:
@redcarrera merging them, but curious why you're converting many of the comments to doxygen.  The creo plugin implementation isn't public API, so we're not going to be generating (or at least shouldn't be generating) doxygen API output for it.  Are we generating doxygen output for it?

Doxygen could be used to document the plugin's implementation, so I assume that's what you were going for and merge accordingly.  I'll note, though, that some of the inline comments that were converted don't apply as you maybe intended.  That is, the "labels" often found on variable declarations like:

```
/** these are important*/
int a;
int b;
int c;

/** more super-important variables */
char *d;
char *e;
```
If you make "these are important" a doxygen comment, it only applies to the 'a' variable where the intent is to document a, b, and c.  That's also a case that really neither should be made into a doxygen comment regardless as they're not documenting API.

Still, like I said, merging because this set of changes shouldn't cause a problem regardless.  Thanks for cleaning up the plugin!

PR #7, opened 2021-11-08
New Material Object Representation

treydinges:
Added new implementation for material objects within BRL-CAD. This new object type is stored in the .g file as a mixture of binary and string data. The items stored by the object struct are as follows:

- bu_vls name
- bu_vls parent (currently awaiting implementation of polymorphism and inheritance)
- bu_vls source
- bu_attribute_value_set physicalProperties
- bu_attribute_value_set mechanicalProperties
- bu_attribute_value_set opticalProperties
- bu_attribute_value_set thermalProperties

I have also added another method within AVS.c/h that is used for the material_describe function. It will get all of the items within the entire AVS as a VLS in a given format (Sean Morrison mentioned there are standards for how sets should be formatted so this might need some tweaking).

The team thought this implementation consisting of multiple AVS's allows for the separation of different types of properties, streamlining the use of material properties throughout the app.

The opticalProperties AVS set specifically is going to be used for our graphics team counterpart in rendering objects with different physical properties.


brlcad (changes requested):
Looking at the direction the code took, bu_avs_get_all() is API-inconsistent.  It's only being called as a helper to print to a string in the describe function.  We don't want to introduce additional public API unnecessarily.  I suggest either changing it to bu_avs_print_vls() and changing the implementation to be consistent with bu_avs_print() or changing the describe function implementation to access the avs directly.

brlcad (approved):
This looks like a really good start to the new material object infrastructure.  Thanks for all your efforts!

PR #8, opened 2021-11-16
Update to art

f4alt:
Update Appleseed art renderer to utilize OSL shaders.
Major changes include how we add items and their respective shaders to the scene:
- We now do this by walking the database
- For now, we assign a precompiled .oso 'disney material' shader to all objects using specified RGB values

Debugging is needed


src/art/art.cpp.old
```diff
@@ -0,0 +1,633 @@
+/*                     A R T . C P P
```
brlcad: Is there a reason for storing a complete copy of art.cpp.old into the repo?  Git has the history if we need to roll back.

src/art/art.cpp
```diff
@@ -271,15 +571,20 @@ asf::auto_release_ptr<asr::Project> build_project(const char* file, const char*
     // Create an empty project.
     asf::auto_release_ptr<asr::Project> project(asr::ProjectFactory::create("test project"));
     project->search_paths().push_back_explicit_path("build/Debug");
+    // ***** add precompiled shaders path
+    // TODO: make this dynamic to the project
+    project->search_paths().push_back_explicit_path("/Users/christopher/Workspace/appleseed/shaders/appleseed");
```
brlcad: should be able to call bu_dir() here to search where appleseed is expected to be installed, or to build up some other path manually (see include/bu/dir.h for examples).

brlcad (approved):
Impressive that you figured this out!

PR #10, opened 2021-12-11
updated gltf importer 

volivarez60:


brlcad (changes requested):
Hi @volivarez60 this commit looks like a mistake or change from earlier.  Can you undo it?  Doesn't look like you submitted a src/libgcv/plugins/gltf plugin so this commit will break the build.

brlcad (changes requested):
@volivarez60 It's probably because you pulled tinygltf from git already. Try deleting the tinygltf/.git folder, then git add src/conv/gltf/tinygltf.  If you do that, you should be able to commit and the PR can be updated with the necessary files.

brlcad (approved):

starseeker:
A thought - can the tinygltf contents be stripped down to a minimum?  (I.e., something like LICENSE, tiny_gltf.* and stb_image* files?)

starseeker:
The examples folder in particular is 23M, and models and tests together add another 1.4M

volivarez60:
Yeah I could get rid of them most likely, I can fix that probably Saturday because im out of town until then.

Get Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

PR #11, opened 2021-12-17
Material command gqa rtweight art

treydinges:
gqa:
- added functionality for new material object representation. Keeps old functionality from .density table but adds the option for material objects to be used in gqa (with or without ids)

rtweight:
- same as gqa for weight calculation
- changed the order of initialization in rt/main.c so that we could use the new material objects from the database

attr:
- Added new standard attribute "material_name" to eventually replace "material_id". Could possibly be renamed to "material" when this change happens

testing:
- unit testing on material command
- integration testing on rtweight
- integration testing on art

Sorry this merge is so big but unfortunately all of the changes were intermingled with each other so I couldn't find a way to separate them through rebasing or cherry-picking commits :(


apstudwell:
Material command:
Import density table materials via id or name into the database via the new material object representation

cdconn00:
Material command: Added command to handle creation, management, and assignment of material properties. Full documentation was added to the command's man page.

brlcad:
Due to a variety of integration issues, this pull request was manually merged into main via command-line.  See commits on 20 Mar 2022 through 6df7001b3482f0bf9e08ee182149ba94086b7851.

PR #16, opened 2022-02-22
Fix various typos

luzpaz:
Found via `codespell -q 3 -L acount,aray,ba,childs,comando,desig,fo,mater,nd,ocur,pard,survice`

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1428,7 +1428,7 @@ s/bu_dirname\(/bu_path_dirname\(/g
 s/bu_basename\(/bu_path_basename\(/g
 s/bu_normalize\(/bu_path_normalize\(/g
 s/bu_argv_from_path\(/bu_path_to_argv\(/g
-        renamed path functons with proper group label [7.27]
+        renamed path functions with proper group label [7.27]
 s/bu_fnmatch\(/bu_path_match\(/g
 s/BU_FNMATCH_LEADING_DIR/0x0/g
 s/BU_FNMATCH_/BU_PATH_MATCH/g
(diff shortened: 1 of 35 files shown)


brlcad:
Hi @luzpaz this looks really impressive, but it can't be applied as-is since you've included modifications to 3rd party codes that are merely co-located in our repository for compilation, and are not ours to modify.  Can you exclude all of src/other and misc/tools?  That will make this a lot easier to review...

brlcad:
Others to exclude include:
doc/docbook/resources/other/docbook5xi.rnc
misc/  (all of it, not just misc/tools)

luzpaz:
@brlcad please recheck

brlcad:
@luzpaz This looks great, thank you!  I'm half-surprised you found as many as you did.  We've had a few devs over the years obsess over spelling.  The ones you found may have been introduced in the past couple years.  I've credited you in our AUTHORS file as Luz Paz <luzpaz>, but let me know if there's another name you'd prefer.

I see from your commit history that you're really good at doing spelling fixes for folks.  Have you considered trying to automate it?  Either on a case-by-case basis, like helping us create a regression test that detects / prevents new mistakes, or across a set of projects where you scan and submit PRs periodically?

luzpaz:
@brlcad I'm happy to do it. You would be surprised at the amount of bugs that have been solved by doing a deeper spell check. I use a combination of [`codespell`](https://github.com/codespell-project/codespell) and vscode regex-fu to run these analyses. As for automation, it's pretty simple to set up a GH action. Just 2 files are needed

`.github/workflows/codespell.yml`
`.github/codespellignore`

Here's how FreeCAD did it:
https://github.com/FreeCAD/FreeCAD/commit/c7cf9b25dfc6ec4fbbba8db7adfd7eb54be105fa

As for automating PRs across several projects, that could be out of my coding scope + codespell does give a number of false positives because it's still a WIP itself. Last thing I want to do is irritate FLOSS devs that are (already) overworked etc...by submitting yet another PR that would not be necessary for them to review. Some devs think spelling for source comments is trivial or that these types of PRs are motivated by an agenda for notoriety. Anyway, like I said, I'm happy to do it as a contribution to FLOSS.

PR #17, opened 2022-02-22
Support to allow to change to quad view at runtime

ddreeves70:
I think I may have commit access but I wasn't 100% sure about this change.  This was more just an educational exercise to get my environment setup and make a small code change. Before I commit to the main I wanted to make sure:

1) I was on the right track with this change.  This is the way to do this.
2) the change was compliant with all the rules for committing code outlined in the HACKING document
3) This was a change we are interested in.

I did some very minor code clean up of the code that was already there.  Mostly I just added the capability to change back and forth to single view and quad view.  There is preexisting code that is not compliant with HACKING document.  If I'm on the right track I will clean up some of the other issues I have seen.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,5 @@
 .DS_Store
 .DS_Store?
 
+.vscode
+.clang-format
--- a/include/qtcad/QtCADQuad.h
+++ b/include/qtcad/QtCADQuad.h
@@ -42,7 +42,7 @@ class QTCAD_EXPORT QtCADQuad : public QWidget
     Q_OBJECT
 
     public:
-	explicit QtCADQuad(QWidget *parent = nullptr, int type = QtCADView_AUTO);
+	explicit QtCADQuad(QWidget *parent, struct ged *gedpRef, int type);
 	~QtCADQuad();
 
 	void stash_hashes(); // Store current dmp and v hash values
@@ -58,12 +58,9 @@ class QTCAD_EXPORT QtCADQuad : public QWidget
 	void select(int quadrant_num);
 	void select(const char *id); // valid inputs: ur, ul, ll and lr
 
-	QtCADView *ur = NULL; // Quadrant 1
-	QtCADView *ul = NULL; // Quadrant 2
-	QtCADView *ll = NULL; // Quadrant 3
-	QtCADView *lr = NULL; // Quadrant 4
-
-	QtCADView *c = NULL;
+        void changeToSingleFrame();
+        void changeToQuadFrame();
+        void changeToResizeableQuadFrame();
 
     signals:
 	void changed();
@@ -75,6 +72,24 @@ class QTCAD_EXPORT QtCADQuad : public QWidget
 
     private:
         bool eventFilter(QObject*, QEvent*);
+        QtCADView *createView(int index);
+        QGridLayout *createLayout();
+
+        int graphicsType = QtCADView_SW;
+        // Holds up to 4 views in single view only the first view is constructed
+        // The other views are constructed if quad view mode is selected by the user
+	QtCADView *views[4] = {nullptr, nullptr, nullptr, nullptr};
+        QtCADView *currentView;
+        // We need to hang on to this because we don't create the other quad views until the user switches mode
+        struct ged *gedp;
+
+        // Hang on to thiese pointers so when we need to clean up memory we have them
+        QGridLayout *currentLayout = nullptr;
+        QSpacerItem *spacerTop = nullptr;
+        QSpacerItem *spacerBottom = nullptr;
+        QSpacerItem *spacerLeft = nullptr;
+        QSpacerItem *spacerRight = nullptr;
+        QSpacerItem *spacerCenter = nullptr;
 };
 
 #endif /* QTCADQUAD_H */
--- a/src/libqtcad/QtCADQuad.cpp
+++ b/src/libqtcad/QtCADQuad.cpp
@@ -43,76 +43,179 @@
 #include <QGridLayout>
 
 #include "bu/str.h"
+#include "ged/defines.h"
+#include "ged/commands.h"
 #include "qtcad/QtCADQuad.h"
 
-QtCADQuad::QtCADQuad(QWidget *parent, int type)
-    : QWidget(parent)
+static const int UPPER_RIGHT = 0;
+static const int UPPER_LEFT = 1;
+static const int LOWER_LEFT = 2;
+static const int LOWER_RIGHT = 3;
+
+static const char *VIEW_NAMES[] = {"Q1", "Q2", "Q3", "Q4"};
+
+/**
+ * @brief Construct a new Qt C A D Quad:: Qt C A D Quad object
+ *
+ * @param parent
+ * @param gedpRef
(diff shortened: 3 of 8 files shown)


starseeker:
@ddreeves70 Taking a look now - impressed you dug into this, it's very experimental code!

starseeker:
This is looking good - the only issue I've noticed is that the embedded raytrace isn't working.  To see this feature in action you can open moss.g with a main build of qged and run the following three commands:

e all.g
view faceplate fb 2
ert

For this to work ert needs some information about the display manager, and in the current pr it's reporting "no current display manager set".  

I think this just means we probably lost a line or two informing the gedp of the display manager somewhere in the refactor, so it should be an easy fix - after that I'd say merge it - nice work!

(Note that ert doesn't quite work correctly with quad-mode even in main - it will always render to the upper right quadrant even if another is selected, which obviously isn't right, so the quad switching code needs to be updating the dm and isn't somewhere - but it does display.  You don't have to fix that in this pull request - although there's certainly no objection to doing so! - just want to be sure that we don't regress ert's behavior.)

ddreeves70:
Ok I'm a little new to this pull request stuff, not sure I did it right.  I
have committed the fix.  BTW I fixed the other issue in your note, just
make sure when you click the other view you issue the 'view faceplate fb 2'
command in addition to the ert command.

starseeker:
Very cool - nice work!

ddreeves70:
Clifford I found another bug I will look at it this weekend. If you select
adc from tools is causing crashes

starseeker:
@ddreeves70 Sounds good - appreciate it!  Out of curiosity, were you looking to make other enhancements to the Qt code?  I've got a laundry list of TODOs to get things up and running on the Qt side, if that's of interest.

starseeker:
@ddreeves70 Probably worth checking to see if any of the more recent changes clear the adc crash - I don't seem to be able to reproduce it here, so I'm not sure, but one or two of them might impact that logic.

ddreeves70:
Yes I plan to continue to work on it I want qged to get to a usable state
which I know is a long ways away. I did review your todo list before I
started, do you have an area you would like me to focus on.

Ok thanks for checking out I probably have a bad compile I have been
experimenting with some stuff.

Seems like you are preferring I take a look at the NURBS stuff. Although I
would like to see a more modernized interface on GED if the NURBS stuff is
more pressing I can take a look at that as well

starseeker:
@ddreeves70 Sean may have a different take, but I would definitely vote for qged - for me at least that is definitely the more pressing concern.

starseeker:
@ddreeves70 nevermind, I see on the zulip chat he agrees.  As far as the adc crash, it's quite possible there's some issue I'm not seeing - either in the view hashing, or some aspect of the interface not fully considered yet.  Linux can be very forgiving on such things, which is occasionally rather maddening when it "works for me" but everybody else is seeing issues.  (This happens a lot with Sean's Mac, for example.)

starseeker:
@ddreeves70 As for areas to focus on... it kinda depends on if you want to start big or small.  My "next step" pieces were to look into what's involved with being able to interactively edit primitives, but that's going to pull in a lot of issues - how to generate and manage updating wireframes for editing without hitting the disk (for example, when altering the radius of a sphere or ell), how to suppress the non-editing wireframes in the view when the editing wireframes come up (MGED doesn't solve that problem well, which is why it will error out if it detects an edit attempt with multiple instances drawn), custom event filters for the editing operations (some of that has been figured out for the polygon editing, but each primitive will likely need its own version), etc.

For a smaller bite, my immediate "next step" was to design a customization of QLineEdit with a QValidator for .g object names, that would check the current text as a db_lookup name and color code it accordingly.  The initial thought for behaviors is coloring it with small stylesheet snippits - green if it matches a "selected" name (either stashed in the widget or pointed to), red if it matches a db object that is not the "selected" object, and orange if it doesn't match anything in the database.  The behavior could switch if the "selected" name is NULL, and highlight green if db_lookup finds a match and red if it doesn't.

starseeker:
I was thinking to define that as a QgLineEdit or some such in libqtcad as a general purpose widget, that would then get used when defining editing widgets and other GUI components that work with .g object names.

starseeker:
A bigger but still somewhat self-contained piece would be to work on translating Archer's sketch editing capabilities into a Qt widget (it doesn't have to be an exact re-implementation of what Archer has, just something that reproduces its capabilities.)

starseeker:
Translating the HUD capabilities into the lower level C libraries out of MGED would probably be good as well, if you want to do that - I've not looked at that much yet, so I don't have a real sense of how difficult it will be.

starseeker:
(for the sketch editing piece, https://github.com/peizhan/psketcher might have some code that would be useful - it's quite an old project now though, so I don't know if it'd be a net win to extract usable pieces from there or start clean.  The advantage of this as a project is the sketch editing widget can be developed independently of the rest of qged in its own small app, and then once the widget it ready it can be moved to libqtcad.)

starseeker:
(if we were going to go for a properly constrained sketch editor I'd probably also go back and investigate https://github.com/cmerrill/sketchsolve but IMHO right now that'd be a distraction from QGED - if we can duplicate what Archer can do that's enough to start with, and once we're over the "replace Archer and MGED" hurdle we'll be well placed to go back and look at those capability-expansion type mods.)

ddreeves70:
I would like to tackle the primitive editing stuff. However, i will do the
QLineEdit first to have one more quick win before I tear into the big one.
If that is cool with you

ddreeves70:
I have been doing some thinking about QLineEdit, I think you are intending
to use this as a field on a form?  If so then I would be a little concerned
about the experience for the user.  Generally speaking it is considered a
bad practice to put lookups or anything like that at the field level.
Normally you want to reserve those types of edits for the form level
edits.  The reason is the performance, the user exits the field trips the
validator and then there is this pause.  Users could be cut and pasting or
half dozen other things they intend to go back and change the field.  Also
I think we want to be careful of our use of color.  It is ok for color to
be a secondary indicator or attention getter but not the primary form of
communication to our user.  Besides the fact some people may be color blind
it is not normally considered overly intuitive.  Maybe an obvious error if
you are using red but may not be obvious as to what type of error has
happened.  Normally you want to leave the field the original color but
maybe color the error message red.


Anyway, I may be way off track on what you are doing if so I apologize.
You may know all this and are making this decision based on some other
reasoning.  Just thought I would put this out there for food for thought.

starseeker:
My intent was to convey whether the name in the field is valid or invalid in the .g database.  I'm not a usability expert, so I'm open to other ideas, but this is the scenario of concern:

I'm wanting to rename a primitive in the tree view or in an editing dialog.  I start typing a new name, but won't know until I press enter whether the name I want to select is already in use in the database.  db_lookup calls are generally quite fast, compared to human typing speed, so it seemed to me we could give immediate feedback on the name on a per-character basis as to what its significance is in the database, rather than having to wait for them to press return.  If there's some better mechanism for indicating an invalid name than color coding that'd be fine too.

If you want to proceed with some other approach on the editing that's fine - I'm not overly tied to the QLineEdit idea, it just struck me as a logical incremental step.

PR #19, opened 2022-03-28
Update tinygltf.dist

redcarrera:
Removed leading space from file name on line 27
Produces error with CMake v3.20.2

### DIFF
--- a/src/conv/gltf/tinygltf.dist
+++ b/src/conv/gltf/tinygltf.dist
@@ -24,7 +24,7 @@ tinygltf/models/box01.glb
 tinygltf/models/CubeImageUriSpaces/2x2 image has spaces.png
 tinygltf/models/CubeImageUriSpaces/CubeImageUriSpaces.bin
 tinygltf/models/CubeImageUriSpaces/CubeImageUriMultipleSpaces.gltf
-tinygltf/models/CubeImageUriSpaces/ 2x2 image  has multiple      spaces.png
+tinygltf/models/CubeImageUriSpaces/2x2 image  has multiple      spaces.png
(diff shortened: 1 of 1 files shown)


brlcad:
@redcarrera That file actually has a space as the first character.  Removing the space makes distcheck fail for me as that file then becomes unaccounted for.  It's not erroring for me.  Can you post what the error you're seeing in cmake is for it?

redcarrera:
> @redcarrera That file actually has a space as the first character. Removing the space makes distcheck fail for me as that file then becomes unaccounted for. It's not erroring for me. Can you post what the error you're seeing in cmake is for it?

Yes, I'll grab the CMake error and provide the exact wording...

redcarrera:
> @redcarrera That file actually has a space as the first character. Removing the space makes distcheck fail for me as that file then becomes unaccounted for. It's not erroring for me. Can you post what the error you're seeing in cmake is for it?

Here's the nastygram:

-- Adding src/conv/gltf...
CMake Error at misc/CMake/BRLCAD_Util.cmake:104 (_message):
  Attempting to ignore non-existent file tinygltf/models/CubeImageUriSpaces/
  2x2 image has multiple spaces.png, in directory
  "C:/Workspace/BRL-CAD/Source/src/conv/gltf"
  
  Result: Configuring incomplete, errors occurred!
  
  I have attached a condensed version of the console log:
  
[BRL-CAD_CMake_Condensed_Console_Log_20220328.txt](https://github.com/BRL-CAD/brlcad/files/8364018/BRL-CAD_CMake_Condensed_Console_Log_20220328.txt)

PR #23, opened 2022-04-04
Remove -GL

GradualSyrup:
Building with CMake 3.22.2 and Visual Studio 2019 Community results in the error:
LINK : fatal error C1047: The object or library file '..\..\..\Debug\lib\tclstub86.lib' was created by a different version of the compiler than other objects like 'Tkhtml.dir\Debug\htmldefaultstyle.obj'; rebuild all objects and libraries with the same compiler

Earlier in the verbose log there's a clue to the issue here:
htmldefaultstyle.obj : MSIL .netmodule or module compiled with /GL found; restarting link with /LTCG; add /LTCG to the link command line to improve linker performance


brlcad:
@starseeker this was for your eyes.  Not sure where those windows-specific build flags came from, but I'm wondering if they can all be eliminated (i.e., let cmake handle it or inherit) since they appear to make assumptions.

starseeker:
IIRC, they come from the original tkhtml build logic.  It would take some study to determine what they do and how they might interact with the Tcl/Tk build's options.

PR #27, opened 2022-04-12
Added z-up fix for art

cameron-norris:
Corrects default rendering behavior of art to correctly render z-up instead of y-up. Might need some refinements but it is functional


brlcad (approved):
More work will be needed for -M arbitrary camera support, but this looks good for getting Z-up working with the azimuth and elevation options.

PR #29, opened 2022-04-19
add art command and button to mged

tanyakejriwal:
Added functionality to call "art" (appleseed rendering) command from mged and added a button on Raytrace Control Panel inside mged to call "art". 

The code calls the command correctly but it does not render with art because there is no -M option for art current. I am working on adding the -M option to art. 

This code allows to call art the same way as we call rt from mged.


brlcad (commented):
There's a couple calls to do_Raytrace that were not updated (one in rt.tcl and another in mouse.tcl) and that's a problem as that appears to break rt.  Documentation is needed as this introduces a new command.  On the whole, this looks good for stubbing in a new 'art' libged command (except file header is wrong).

PR #30, opened 2022-04-19
Art Options

acurtis55:
This version of art.cpp has [o, s, w, C, n, A, a, e, p, E ] options working while running art. The functionalities are equivalent to that of rt.

### DIFF
--- a/src/art/art.cpp
+++ b/src/art/art.cpp
@@ -17,7 +17,25 @@
  * License along with this file; see the file named COPYING for more
  * information.
  */
-/*********************************************************************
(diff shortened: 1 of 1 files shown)


brlcad (changes requested):
There are a number of issues in this PR.  It looks like you did a git pull at some point, but then overwrote the current file with an older edited version that you were working on.  This wiped out other people's work, so the PR can't be committed.

It also looks like your editor changed the indentation on many lines of code so they show up as edited lines also, making it hard to figure out what you actually changed.  Please submit a clean PR with just your intended changes.

PR #38, opened 2022-05-03
gcv glTF geometry read

atp318:
-adds ability for gcv to read geometry files with glTF geometry
-accepts ".gltf" and ".glb" as input file types
-"/tinygltf_headers/*" are the necessary header files to use tinyglTF, a header library used to read the object files
-creates object geometry for use in BRL-CAD, outputs ".g" file type
-imports object names, relationships, hierarchies, and transformations


brlcad (approved):
This looks like good incremental progress on gltf support for gcv.  Exhaustive testing isn't warranted since this is still experimental (and still likely need to keep the src/conv variant for reference a bit longer), but it does appear to be handling entities properly in a nice procedural structure.

PR #39, opened 2022-05-03
Creo to BRL-CAD Converter as of 4/20/2022

redcarrera:
These are the bulk changes to the converter for
compatibility with Creo version 7.0.6.0

### DIFF
--- a/assembly.cpp
+++ b/assembly.cpp
@@ -1,0 +1,374 @@
+/**
+ *                  A S S E M B L Y . C P P
+ * BRL-CAD
+ *
+ * Copyright (c) 2017-2022 United States Government as represented by
+ * the U.S. Army Research Laboratory.
+ *
+ * This program is free software; you can redistribute it and/or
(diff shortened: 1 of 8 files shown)


brlcad:
Hey @redcarrera Looks like something went wrong with the git add and git push to your repo.  The files appear to be added to the top-level of the repo instead of showing up as edits to the existing files.  You'll probably want to start with a fresh branch (or force undo the one commit), make your changes/edits to the files where they reside in src/external/Creo, then "git diff" to make sure the only changes are the ones you intended to make.  If that diff looks good, you run "git add src/external/Creo" in your branch and can run "git diff --staged" again to make sure it's still only what you intend to commit.  Then can run "git status" to make sure you didn't miss anything, and then commit.  Once committed locally, you can git push (or git push --force to update this same PR branch).

brlcad (changes requested):
Adds files to top-level and should not.  Should be edits to files in src/external/Creo only.

PR #43, opened 2022-05-08
art button and command in mged

tanyakejriwal:
Added art button and command in mged. Made following changes:
1. Deleted some extra code and cleaned it up
2. calling rtabort or pressing "abort" will abort art
3. help art outputs art help message
4. fixed do_Raytrace calls so that rt doesn't break


brlcad (approved):
usage still says "art test" but no the whole this looks really good.  implements -M as well as the GUI changes for adding an art button (which of course uses -M).

Issue #28, opened 2022-04-19
Wrong computation of the moments of inertia


drossberg:
[cubes.zip](https://github.com/BRL-CAD/brlcad/files/8513608/cubes.zip) contains two files with cubes with 1m side length (plus density data). One cube is centered at the origin, the other 10 m away.

The output for `gqa -g 1 centered_cube.g all.g` contains
```
MATRIX For the Moments and Products of Inertia For all.g:
   166080893923271.375 -84081549.121 -84081549.121    0.000
   -84081549.121 166080893923270.375 -84081549.121    0.000
   -84081549.121 -84081549.121 166080893923271.688    0.000
      0.000    0.000    0.000    0.000
```

The output for `gqa -g 1 eccentered_cube.g all.g` contains
```
MATRIX For the Moments and Products of Inertia For all.g:
   166080893923270.250 9185826187.530 9185826187.576    0.000
   9185826187.530 -42984348272864.000 -84081549.121    0.000
   9185826187.576 -84081549.121 -42984348303264.000    0.000
      0.000    0.000    0.000    0.000
```

The moments of inertia should however never be negative.

BTW, `gqa` without the `-g` option seems to never run infinitely.

drossberg:
A pull request is on the way.

drossberg:
After applying the #44 I get:
```
MATRIX For the Moments and Products of Inertia For all.g:
   166667384642965.500 27777777.778 27777777.778    0.000
   27777777.778 166667384642965.500 27777777.778    0.000
   27777777.778 27777777.778 166667384642965.062    0.000
      0.000    0.000    0.000    0.000
```
and
```
MATRIX For the Moments and Products of Inertia For all.g:
   166667384642965.312 27777777.777 27777777.735    0.000
   27777777.777 166667384950112.000 27777777.778    0.000
   27777777.735 27777777.778 166667384919456.000    0.000
      0.000    0.000    0.000    0.000
```
Looks better, I would say.

PR #44, opened 2022-05-12
consistent grid spacing

drossberg:
- always use gridSpacing as grid spacing value
- shifted the grid a bit more to the middle by starting the first ray at "min + gridSpacing / 2" instead of "min + gridSpacing"
- the rays are generated from 1 to state.steps - 1, i.e. the number of steps is in fact state.steps - 1, which should be state.span / gridSpacing, on the other hand


drossberg:
This should solve #28.

drossberg:
Committed the patch after the discussion on Zulip.

PR #47, opened 2022-06-02
Nirt options cleanup when run within mged

f4alt:
nirt help is now useful and prints using src/nirt. The fix applies to all commands which solely print information to the users (ie -L)
Some work on -f fmt is wrapped in as well. Option is now recognized, but needs format cleanup.


brlcad (approved):
This all looks really good to me; don't see any issues.

PR #50, opened 2022-06-20
Gqa path

f4alt:
Resolve the crash when gqa is supplied a path with slashes


starseeker (commented):
Question - with this change, what does gqa do if handed an object whose name contains a slash?  It's bad practice, but it can happen...  (for example, a sphere created with the MGED command: make "test.s/sph" sph)

PR #54, opened 2022-07-14
connect asset importer library to gcv

f4alt:
add assimp API to gcv reader and writer to allow conversion between all assimp supported import/export filetypes and .g databases.

Relies on building with "Assimp_ROOT=/path/to/assimp" and "Assimp_ENABLED=ON"


starseeker:
@brlcad Can this be merged in without causing you problems?

PR #59, opened 2022-09-25
Add new rtdepth program

erikg:
Proposal rt program for depth computation. Useful for 3d cnc milling.


brlcad:
This looks good to me.  I thought we had a tool that did this, but the closest I could find was rtxray, which is cumulative an inverted, which you wouldn't really want.

The only issues I see are that it's got different indent and no docs.. Can you write up a man page for it?

PR #71, opened 2023-04-01
removed the "d" in MS Windows Debug binaries

drossberg:
for 3rd party libraries
they were in the way during post-processing and prevent the debug build in some common constellations


starseeker:
So, do I understand correctly that when you're building in debug mode, the output dll file names on your system are NOT showing the "d" naming convention?

drossberg:
No, the binary debug output files have the "d" in their names, but the consuming libraries expect them without the "d".

When I generate the build files with standard CMake for MS Visual Studio, CMAKE_BUILD_TYPE is an empty string by default, and a solution is generated, which contains a configuration for release and debug mode. The scripts in src/other/ext however add the "d" to the set-up for the BRL-CAD libraries if CMAKE_BUILD_TYPE is set (and something like "debug"). As a result, the libraries with the "d" won't be found in debug mode, because the consuming libraries expect them without the "d".

In principle, this could be fixed by using a construct like `$<$<CONFIG:debug>:~`. But unfortunately, it won't be expanded for all occurrences.

PR #75, opened 2023-04-27
function to list available read and write formats in libgcv

nishanth-01:


brlcad:
Why was this closed?

nishanth-01:
Working on a better one

PR #82, opened 2023-07-14
Creo to BRL-CAD Converter as of 7/14/2023

redcarrera:
Enclosed are the bulk changes to the converter for use with Creo version 8.0.8.0


starseeker (changes requested):
@redcarrera Can you try building this on a Linux machine as part of the BRL-CAD build?  I'm seeing some errors about unused variables and things like signed/unsigned comparisons.

redcarrera:
@starseeker  Sorry, I do not currently have access to a Linux machine.  If you share the
 mcc compiler message log, I would gladly chase down and eliminate the unused variables
 as mentioned above

PR #89, opened 2023-07-26
Creating basic type command for VDB

cjsb:
Creating the basic structure to load a new type on the mged project.

If someone runs mged and do the command "make vdbName vdb" it will run the code correctly but the ray tracing function it is missing and an error will appears. The program does not break.


erikg:
Changed merge target to BRL-CAD:vdb, this shouldn't merge to main until the primitive is complete.

This commit changes a lot of indentation unnecessarily, please fix that

cjsb:
Thank you for adding the code in the correct branch.

My vdb is only a bounding box at the moment to make it easier for me to understand if the data is at the correct place. The ray intersection calculation is getting the data correctly and computing the intersection. Next step might be adding a actual vdb now.

include/bu/magic.h
```diff
@@ -125,6 +125,7 @@ __BEGIN_DECLS
 #define RT_TGC_INTERNAL_MAGIC		0xaabbdd87 /**< ???? */
 #define RT_TOR_INTERNAL_MAGIC		0x9bffed87 /**< ???? */
 #define RT_VOL_INTERNAL_MAGIC		0x987ba1d0 /**< ?{?? */
+#define RT_VDB_INTERNAL_MAGIC		0xa7a3d52a /**< vdb */
```
H1manshu21, line 128: Since all the macro name corresponds to ASCII representation of the hex numbers which gives a hint of what they means. For e.g. below `706e7473` means `pnts`. So similarly you can change `vdb` to `76646220`.

src/libged/garbage_collect/garbage_collect.cpp
```diff
@@ -30,6 +30,7 @@
 #include <set>
 #include <string>
 #include <vector>
+#include <iterator>
```
H1manshu21, line 33: `set` library already has member type `iterator` that is used here in line 76 `std::set<std::string>::iterator s_it;`. So no need to include `iterator` library.

H1manshu21 (changes requested):

include/bu/magic.h
cjsb: Cool! I will do that.

src/libged/garbage_collect/garbage_collect.cpp
cjsb: I did that because I was getting an error on line 261 and 262: error C2039: 'inserter': is not a member of 'std'.

That was fixed by adding the include.

PR #93, opened 2023-08-04
Fix a variety of typos in the manual

joshbaker22:


brlcad (approved):
These all look like great changes to me.

PR #95, opened 2023-08-17
Updates to trimesh.h and trimesh.cpp

redcarrera:
Clarified role of function bg_trimesh_solid:
Check if a mesh is topologically solid. Returns TRUE if the mesh is
NOT SOLID and FALSE if the mesh is SOLID. A FALSE outcome indicates
the mesh satisfies all three criteria:  Closed, Manifold, Oriented


starseeker (approved):
Looks good

PR #97, opened 2023-08-24
Nurbs editing Milestone 3

GregoryLi0:
1. Rearrange NURBS module into geometry parts and topology parts.
2. Simple line editing operations in 2D parameter space.
3. Support for topology edges, faces, loops, and trimmes.


drossberg:
In src/libged/brep/geometry.cpp there are some strange tabulators in the printed text. E.g. in lines 68 and 97.

GregoryLi0:
Hi, I just fixed them in geometry.cpp and topology.cpp

Issue #100, opened 2023-10-25
unable to suppress bu_log from multiple threads simultaneously


kennethsdavis:
I am using the BRL-CAD API in a multi-threaded application and would like to suppress all log messages printed to the screen by bu_log. Calling bu_log_hook_delete_all and bu_log_add_hook with an empty function is a near solution. However, if one thread is inside the call to log_call_hooks when a second thread calls bu_log, then the second thread will have its message printed to the screen.

Providing a function that silences bu_log would solve the problem. Though it might be nice to make bu_hook_call thread-safe.

BRL-CAD 7.32.2

kennethsdavis:
To reproduce the error on windows and linux, build and run the following code.
```
#include <algorithm>
#include <thread>
#include <vector>

#include <brlcad.h>

void thread_function()
{
   while(true)
   {
      bu_log("bu_log: thread log message\n");
      std::this_thread::sleep_for(std::chrono::milliseconds(10));
   }
};

int main(int argc, char** argv)
{
   bu_log("bu_log: test log message\n");

   // after this call to bu_log_add_hook no calls to bu_log should result in a
   // message being printed
   bu_log_add_hook([](void*,void*) { return 0; }, NULL);

   std::vector<std::thread> threads;

   // set thread_count to 1 to verify that bu_log doesn't print any messages in
   // a single-threaded context
   const unsigned int thread_count = std::thread::hardware_concurrency();
   std::generate_n(std::back_inserter(threads), thread_count, []() { return std::thread(thread_function); });

   // because the thread_function has an infinite loop the only way to quit is
   // to CTRL-C your way out of the application. all remaining code is
   // effectively useless.
   std::for_each(std::begin(threads), std::end(threads), [](auto& thread) { thread.join(); });

   return EXIT_SUCCESS;
}
```

Walker0710:
hello i'm a beginner, can you please assign me some beginner level issues or any bugs.

brlcad:
@AnkushSingh0710 Since you're totally commenting on this thread seemingly randomly -- you could try to address this issue.

@kennethsdavis Thank you for the report.  I'd be very interested to hear more about this multithreaded application of yours.  That said, thank you very much for the code snippet that demonstrates this issue and the detailed writeup.  I agree that pattern should work the way you're calling without erroneous logging and is quite unexpected.  I'm guessing we don't see it in our own applications because logging is not suppressed by default.  There certainly isn't a unit test for it (as yet).

[...]

Issue #105, opened 2023-12-12
brl-cad dll build errors


kennethsdavis:
7.36.0
- There are a number of unresolved dependencies on functions in libbv. Add libbv-static to the target_link_libraries
- libopennurbs.lib has an unresolved dependency on PathFileExists (as well as a couple other related functions). Add shlwapi to COREINTERFACE_MSVC_LIBRARIES

[7-36-0_brlcad-dll.patch](https://github.com/BRL-CAD/brlcad/files/13650349/7-36-0_brlcad-dll.patch)

7.38.1
- Link step of libbg-static.lib into brlcad.dll results in many symbols already defined in libbv-static.lib (vlist.obj).
- nmg_booltree_leaf_tess referenced in brlcad.def doesn't exist in the brlcad library.
-- correction, it exists but ConstDatabase.cpp in rt^3 can't find it via raytrace.h for some reason

kennethsdavis:
I don't know what the solutions are to the 7.38.1 errors. If the symbols are removed from either libbg-static or libbv-static, then other targets fail to build.

starseeker:
I applied your patch and also fixed a couple missed function name adjustments (commit 03ffae5bfd) - for the vlist.obj problem, can you provide a couple examples of the errors you're seeing?

kennethsdavis:
@starseeker, I rebased my main branch, but ConstDatabase.cpp in rt-cubed still has a reference to nmg_booltree_leaf_tess. Replacing that with rt_booltree_leaf_tess allows me to get to the link phase. Here are the duplicate symbols you asked for:
```
20>------ Build started: Project: brlcad, Configuration: Release x64 ------
20>ConstDatabase.cpp
20>libbg-static.lib(polygon.obj) : error LNK2005: pd already defined in libbv-static.lib(vlist.obj)
20>libbg-static.lib(polygon.obj) : error LNK2005: pd_3 already defined in libbv-static.lib(vlist.obj)
20>libbg-static.lib(polygon.obj) : error LNK2005: pd_3box already defined in libbv-static.lib(vlist.obj)
20>libbg-static.lib(polygon.obj) : error LNK2005: pd_3cont already defined in libbv-static.lib(vlist.obj)
20>libbg-static.lib(polygon.obj) : error LNK2005: pd_3line already defined in libbv-static.lib(vlist.obj)
20>libbg-static.lib(polygon.obj) : error LNK2005: pd_3move already defined in libbv-static.lib(vlist.obj)
[...]

PR #106, opened 2023-12-12
suppress bu_log from multiple threads simultaneously

kennethsdavis:
There is a semaphore global to log.c that is intended to protect calls to bu_hook_call (via log_call_hooks). Because log_call_hooks can only be called if the log_hook_list is non-empty this semaphore only gets initialized when bu_log_add_hook is called the first time.

I considered the FIXME note at the top of log.c as a strong candidate for implementation from a design standpoint, but favored this solution because its development cost and risk are both low.

This pull request satisfies https://github.com/BRL-CAD/brlcad/issues/100


kennethsdavis:
Testing was performed against 7.36.0 because main HEAD won't build brlcad.dll at the time of this comment (see https://github.com/BRL-CAD/brlcad/issues/105). The sample application on https://github.com/BRL-CAD/brlcad/issues/100 executes as expected after applying this change.

kennethsdavis:
I may be getting close to updating my brl-cad version. It would be nice to have this fix in the update. Is there anything I can do to help in the review of this PR?

starseeker:
I think this should be OK - I'll go ahead and merge.  @brlcad let me know if you see any issues here.

PR #108, opened 2023-12-29
rt_regionfix polite error reporting

### LINKED ISSUE #101: need option to disable rt_regionfix
Using the API, it would be nice to prevent rt_prep_parallel from calling rt_regionfix. The default behavior of rt_regionfix is to look for a model.regexp file in the same folder as model.g and use it to re-number the regions defined in model.g. This feature allows users to modify developer-delivered models to affect application output.

kennethsdavis:
rt_regionfix was treating the inability to open a region fix file as an error worthy of direct printing to stderr. However this condition didn't invalidate the rt_i and the calling function (rt_prep_parallel) remained blissfully unaware.

After this fix rt_regionfix calls bu_log instead of perror. This permits error reporting on the user's terms: if the log is to be saved to file, then so be it; if the log is to be ignored, then so be it; etc.

fixes #101

### DIFF
--- a/src/librt/regionfix.c
+++ b/src/librt/regionfix.c
@@ -71,7 +71,7 @@ rt_regionfix(struct rt_i *rtip)
 
     fp = fopen(file, "rb");
     if (fp == NULL) {
-	if (rtip->rti_region_fix_file) bu_log("%s: unable to open REGEXP file\n", file);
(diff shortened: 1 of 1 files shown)


kennethsdavis:
Is there anything I can do to help in the review of this PR?

Issue #116, opened 2024-03-14
[windows]complie with appleseed error


Rainy-fall-end:
I have compiled BRL-CAD from source checkout, but when I compiled 'art' by following [wiki](https://brlcad.org/wiki/Appleseed), something wrong occurs
ENV:visual studio 2022, appleseed 2.1.0-beta
It seems that ArtTileCallback is a virtual class and it can't be constructed directly.
Do I need to replace an earlier version for appleseed?
<img width="527" alt="pic" src="https://github.com/BRL-CAD/brlcad/assets/99249751/b95db996-17fa-4007-b625-368bdf5bcae2">

PR #117, opened 2024-03-21
Fix appleseed error when appleseed version > 2.0 #116

Rainy-fall-end:
update ArtTileCallback class according to Appleseed 1.7+


Rainy-fall-end:
In appleseed [master branch](https://github.com/appleseedhq/appleseed/blob/master/src/appleseed/renderer/kernel/rendering/itilecallback.h),it declares that:
```cpp
class APPLESEED_DLLSYMBOL ITileCallback
  : public foundation::IUnknown
{
virtual void on_tile_begin(
        const Frame*            frame,
        const size_t            tile_x,
        const size_t            tile_y,
        const size_t            thread_index,
        const size_t            thread_count) = 0;
virtual void on_progressive_frame_update(
        const Frame&            frame,
        const double            time,
        const std::uint64_t     samples,
        const double            samples_per_pixel,
        const std::uint64_t     samples_per_second) = 0;
};
}
```
In [appleseed 2.0+](https://github.com/appleseedhq/appleseed/blob/2.0.5-beta-maintenance/src/appleseed/renderer/kernel/rendering/itilecallback.h)(2.0.5 beta and 2.1.0 beta)：
```cpp
class APPLESEED_DLLSYMBOL ITileCallback
  : public foundation::IUnknown
{
virtual void on_tile_begin(
        const Frame*    frame,
        const size_t    tile_x,
        const size_t    tile_y) = 0;
virtual void on_progressive_frame_update(const Frame* frame) = 0;
}
```

Rainy-fall-end:
I I simply updated the methods from 1.0+ to 2.0+, I'm not sure if this is still appropriate or if it's necessary to support all appleseed versions.

drossberg:
Just some formal remarks:
* Do not comment out old code, just change what need to be changed. The version control system (i.e. git) does the bookkeeping, as you can see the the "Files changed" tab.
* You should remove the clutter from your PR ;)

Rainy-fall-end:
> Just some formal remarks:
> 
> * Do not comment out old code, just change what need to be changed. The version control system (i.e. git) does the bookkeeping, as you can see the the "Files changed" tab.
> * You should remove the clutter from your PR ;)

Thanks for your comments. I have deleted it all.

PR #121, opened 2024-03-31
rtsurf compute volume 

bralani:
The main task of this PR is to compute the volume of the 3D object using Monte Carlo sampling in rtsurf module.

I split the code in the bounding_sphere_sampling function because it will be useful in the future for other tasks (like sampling in neural rendering).

N.B.: Do not merge because the PR is not complete!

Papers related: 
- Montecarlo Methods to Efficently Compute Volume of Solids 
- A Monte Carlo Method for High-Dimensional Volume Estimation
and Application to Polytopes


brlcad:
Thanks @bralani this looks like steps in the right direction.  Do you plan to complete this change, hooking it through to an rtsurf command-line option or other output?  Would be good to see at least a basic validation test case as well to prove it's working.

bralani:
@brlcad I plan **at least** to implement by command-line option. Yes, of course I will do some validation tests.

bralani:
I implemented the function to estimate the volume of an object using `rt_shootray_simple` instead of `rt_shootray` because it returns the list of partitions of the ray (list of segments of the ray inside the object). Then, the volume of the object is calculated similarly to the surface with the following formula:

$V = \frac{4}{3} \cdot \pi \cdot r^3 \cdot \frac{segs_{\text{obj}}}{segs_{\text{bs}}}$

where:
- $segs_{\text{obj}}$ is the overall length of the segments which intersect the object.
- $segs_{\text{bs}}$ is the overall length of the segments inside the bounding sphere (rays generated by Monte Carlo sampling).

If you want to try it, simply put the `-v` option in the terminal.

Here there are two samples, one with a cube of edge=2000mm:
$\text{S} = 6 \times (\text{edge})^2 = 24000000 \ \text{mm}^2$
$\text{V} = (\text{edge})^3 = 8000000000 \ \text{mm}^3$

image

And one with a sphere of radius=1000mm:
$\text{S} = 4\pi r^2\ = 12566370,61 \ \text{mm}^2$
$\text{V} = \frac{4}{3} \pi r^3 = 4188790205 \ \text{mm}^3$
 
image

@brlcad How can I create a basic validation test case? I mean, where should I create it...

bralani:
There is a minor graphical issue related to the output, for some reason it prints always 0 as the total length segments but the estimated volume is right. I will resolve it as soon as possible.
Screenshot 2024-04-04 alle 21 00 50

bralani:
I resolved the graphical issue, I need to create some validation tests. Any suggestion is welcome.

PR #127, opened 2024-04-20
LS-Dyna keyword file to BRL-CAD converter

drossberg:
replaced the old implementation by a more feature-rich one


starseeker:
@drossberg Looking over the existing k-g, it looks like it has support for ELEMENT_SOLID import to arb primitives.  Does this new version still support that?  I'm not seeing any ELEMENT_SOLID logic in a quick look.

PR #140, opened 2024-06-06
Handling unexpected command *PART_ADAPTIVE_FAILURE

AliHaider93:
this code is written to handle unexpected command *PART_ADAPTIVE_FAILURE by adding it as an attribute to the region.

### DIFF
--- a/src/conv/k-g/k-g.cpp
+++ b/src/conv/k-g/k-g.cpp
@@ -122,6 +122,9 @@ int main
 				bot.addTriangle(point1, point3, point4);
 			    }
 			}
+			if ((it->second).attributes.size() > 0) {
+			    regions.setAttributes(partName,(it->second).attributes);
+			}
 		    }
 		}
 
--- a/src/conv/k-g/k_parser.cpp
+++ b/src/conv/k-g/k_parser.cpp
@@ -39,7 +39,8 @@ enum class KState {
     Node,
     Element_Shell,
     Part,
-    Section_Shell
+    Section_Shell,
+    PART_ADAPTIVE_FAILURE
 };
 
 
@@ -188,6 +189,9 @@ bool parse_k
 				state         = KState::Part;
 				partLinesRead = 0;
 				partTitle     = "";
+			    }
+			    else if ((command.size() == 3)) {
+				state = KState::PART_ADAPTIVE_FAILURE;
 			    }
 			    else
 				std::cout << "Unexpected command " << tokens[0] << " in k-file " << fileName << "\n";
@@ -346,6 +350,22 @@ bool parse_k
 			}
 
 			++sectionLinesRead;
+			break;
+		    }
+		    case KState::PART_ADAPTIVE_FAILURE: {
+			if (tokens.size() == 0) {
+			    std::cout << "Too short PART_ADAPTIVE_FAILURE in k-file" << fileName << "\n";
+			    break;
+			}
+			int pid = stoi(tokens[0]);
+
+			if (tokens.size() == 2) {
+			    data.parts[pid].attributes["ADAPTIVE_FAILURE_STARTING_TIME"]=tokens[1];
+			}
+			else {
+			    std::cout << "Attributes should come in pairs in k-file" << fileName << "\n";
+			    break;
+			}
 			break;
 		    }
 
--- a/src/conv/k-g/region_list.cpp
+++ b/src/conv/k-g/region_list.cpp
@@ -39,13 +39,48 @@ Bot& RegionList::addRegion
 (
     const std::string& name
 ) {
-    Bot& ret = m_list[name];
+    Bot& ret = m_list[name].bot;
 
     std::string botName(name);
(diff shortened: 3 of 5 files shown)


src/conv/k-g/k-g.cpp
```diff
@@ -122,6 +122,9 @@ int main
 				bot.addTriangle(point1, point3, point4);
 			    }
 			}
+			if ((it->second).attributes.size() > 0) {
```
drossberg, line 125: You do not need this test here. The behavior stays the same without it.

src/conv/k-g/k_parser.cpp
```diff
@@ -39,7 +39,8 @@ enum class KState {
     Node,
     Element_Shell,
     Part,
-    Section_Shell
+    Section_Shell,
+    PART_ADAPTIVE_FAILURE
```
drossberg, line 43: Please, follow the style of the other states, i.e. write *Part_Adaptive_Failure* and put it below *Part*. It doesn't matter, if the states below getting other numbers.

src/conv/k-g/k_parser.cpp
```diff
@@ -189,6 +190,9 @@ bool parse_k
 				partLinesRead = 0;
 				partTitle     = "";
 			    }
+			    else if ((command.size() == 3)) {
```
drossberg, line 193: Test the whole string: `else if ((command.size() == 3) && (command[1] == "ADAPTIVE") && (command[2] == "FAILURE")) {`

src/conv/k-g/k_parser.cpp
```diff
@@ -348,6 +352,22 @@ bool parse_k
 			++sectionLinesRead;
 			break;
 		    }
+		    case KState::PART_ADAPTIVE_FAILURE: {
+			if (tokens.size() == 0) {
```
drossberg, line 356: There are a bit to many tests for *tokens.size()* here ;) One `if (tokens.size() < 2) {` then error is enough. If there are more than 2, the surplus will be ignores, as in the other states.

src/conv/k-g/k_parser.cpp
```diff
@@ -348,6 +352,22 @@ bool parse_k
...
 		    }
+		    case KState::PART_ADAPTIVE_FAILURE: {
+			if (tokens.size() == 0) {
+			    std::cout << "Too short PART_ADAPTIVE_FAILURE in k-file" << fileName << "\n";
+			    break;
+			}
+			int pid = stoi(tokens[0]);
+
+			if (tokens.size() == 2) {
+			    data.parts[pid].attributes["ADAPTIVE_FAILURE_STARTING_TIME"]=tokens[1];
```
drossberg, line 363: *PART_ADAPTIVE_FAILURE_STARTING_TIME*?

src/conv/k-g/region_list.cpp
```diff
@@ -49,16 +49,53 @@ Bot& RegionList::addRegion
...
     BU_LIST_INIT(&all_head.l);
 
-    for (std::map<std::string, Bot>::iterator it = m_list.begin(); it != m_list.end(); ++it) {
-	const std::string& region_name = it->first;
-	Bot&               bot         = it->second;
+    for (std::map<std::string, Region>::iterator it = m_list.begin(); it != m_list.end(); ++it) {
+	const std::string&                 region_name       = it->first;
+	Region&                            region            = it->second;
+	Bot&                               bot               = region.bot;
+	std::map<std::string, std::string> region_attributes = region.attributes;
```
drossberg, line 98: Const reference?

src/conv/k-g/region_list.h
```diff
@@ -32,24 +32,36 @@
...
 
 class RegionList {
 public:
     RegionList(void);
 
     Bot& addRegion(const std::string& name);
 
-    Bot& getBot(const std::string& name);
+    Bot&                                getBot(const std::string& name);
+    std::map<std::string, std::string>& getAttributes(const std::string& name) ;
```
drossberg, line 47: It's nowhere needed, therefore I would recommend to omit this.

src/conv/k-g/region_list.h
```diff
@@ -32,24 +32,36 @@
...
 
     void deleteRegion(const std::string& name);
 
+    void setAttributes(const std::string& name,
+		       const std::map<std::string, std::string>& attributes);
+
+    void writeAttributes(rt_wdb* wdbp,
```
drossberg, line 54: I would make this function at least private. But I think, it can be made as a simple static function in region_list.cpp, without mentioning it in the class declaration at all.

src/conv/k-g/k_parser.cpp
AliHaider93: the attribute Part_Adaptive_Failure in LS-DYNA refers to the time when the adaptive mesh should start. we could call it ADAPTIVE_FAILURE_MESH_STARTING_TIME. I am open for any other suggestions

PR #142, opened 2024-06-06
Handling unexpected command *PART_ADAPTIVE_FAILURE

AliHaider93:
This code is written to handle unexpected command *PART_ADAPTIVE_FAILURE by adding it as an attribute to the region.

### DIFF
--- a/src/conv/k-g/k-g.cpp
+++ b/src/conv/k-g/k-g.cpp
@@ -123,7 +123,7 @@ int main
 			    }
 			}
 
-			    regions.setAttributes(partName,(it->second).attributes);
+			regions.setAttributes(partName,(it->second).attributes);
 		    }
 		}
 
--- a/src/conv/k-g/k-g.cpp
+++ b/src/conv/k-g/k-g.cpp
@@ -122,6 +122,8 @@ int main
 				bot.addTriangle(point1, point3, point4);
 			    }
 			}
+
+			    regions.setAttributes(partName,(it->second).attributes);
 		    }
 		}
 
--- a/src/conv/k-g/k_parser.cpp
+++ b/src/conv/k-g/k_parser.cpp
@@ -353,6 +353,11 @@ bool parse_k
 			break;
 		    }
 		    case KState::Part_Adaptive_Failure: {
+			if (tokens.size() < 2) {
+			    std::cout << "Too short PART_ADAPTIVE_FAILURE in k-file " << fileName << "\n";
+			    break;
+			}
+
 			int pid = stoi(tokens[0]);
 
 			if (tokens.size() == 2) {
@@ -360,7 +365,6 @@ bool parse_k
 			}
 			else {
 			    std::cout << "Attributes should come in pairs in k-file" << fileName << "\n";
-			    break;
 			}
 			break;
 		    }
--- a/src/conv/k-g/k_parser.cpp
+++ b/src/conv/k-g/k_parser.cpp
@@ -39,6 +39,7 @@ enum class KState {
     Node,
     Element_Shell,
     Part,
+    Part_Adaptive_Failure,
     Section_Shell
 };
 
@@ -188,6 +189,9 @@ bool parse_k
 				state         = KState::Part;
 				partLinesRead = 0;
 				partTitle     = "";
+			    }
+			    else if ((command.size() == 3) && (command[1]=="ADAPTIVE") && (command[2]=="FAILURE")) {
+				state = KState::Part_Adaptive_Failure;
 			    }
 			    else
 				std::cout << "Unexpected command " << tokens[0] << " in k-file " << fileName << "\n";
@@ -346,6 +350,18 @@ bool parse_k
 			}
 
 			++sectionLinesRead;
+			break;
+		    }
+		    case KState::Part_Adaptive_Failure: {
+			int pid = stoi(tokens[0]);
+
+			if (tokens.size() == 2) {
+			    data.parts[pid].attributes["PART_ADAPTIVE_FAILURE_STARTING_TIME"]=tokens[1];
+			}
+			else {
+			    std::cout << "Attributes should come in pairs in k-file" << fileName << "\n";
+			    break;
+			}
 			break;
 		    }
 
--- a/src/conv/k-g/region_list.cpp
+++ b/src/conv/k-g/region_list.cpp
@@ -74,17 +74,19 @@ static void  writeAttributes
 	return;
     }
 
-    if (rt_db_get_internal(&region_internal, dp, dbip, NULL, &rt_uniresource) <= 0) {
-	bu_log("writeAttributes() rt_db_get_internal(%s) FAIL, Can't write attributes\n", dp->d_namep);
-	return;
+    if (rt_db_get_internal(&region_internal, dp, dbip, NULL, &rt_uniresource) > 0) {
+	bu_attribute_value_set* avs = &region_internal.idb_avs;
+
+	for (std::map<std::string, std::string>::const_iterator it = attributes.begin(); it != attributes.end(); it++) {
+	    bu_avs_add(avs, it->first.c_str(), it->second.c_str());
+	}
+
+	db5_update_attributes(dp, avs, dbip);
     }
-    bu_attribute_value_set* avs = &region_internal.idb_avs;
-    for (std::map<std::string, std::string>::const_iterator it = attributes.begin(); it != attributes.end(); it++)
-    {
-	bu_avs_add(avs, it->first.c_str(), it->second.c_str());
+    else {
+	bu_log("writeAttributes() rt_db_get_internal(%s) FAIL, Can't write attributes\n", dp->d_namep);
     }
 
-    db5_update_attributes(dp, avs, dbip);
     rt_db_free_internal(&region_internal);
 }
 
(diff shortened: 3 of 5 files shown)


src/conv/k-g/k-g.cpp
```diff
@@ -122,6 +122,8 @@ int main
 				bot.addTriangle(point1, point3, point4);
 			    }
 			}
+
+			    regions.setAttributes(partName,(it->second).attributes);
```
drossberg: This has to be on the same level as the closing bracket, i.e. 4 spaces back.

src/conv/k-g/k_parser.cpp
```diff
@@ -348,6 +352,18 @@ bool parse_k
 			++sectionLinesRead;
 			break;
 		    }
+		    case KState::Part_Adaptive_Failure: {
+			int pid = stoi(tokens[0]);
```
drossberg, line 361: Check, if there is a token, first.

src/conv/k-g/k_parser.cpp
```diff
@@ -348,6 +352,18 @@ bool parse_k
...
 		    }
+		    case KState::Part_Adaptive_Failure: {
+			int pid = stoi(tokens[0]);
+
+			if (tokens.size() == 2) {
+			    data.parts[pid].attributes["PART_ADAPTIVE_FAILURE_STARTING_TIME"]=tokens[1];
+			}
+			else {
+			    std::cout << "Attributes should come in pairs in k-file" << fileName << "\n";
+			    break;
```
drossberg: The "break" is not needed here. The next one will do it for all.

src/conv/k-g/region_list.cpp
```diff
@@ -49,16 +49,53 @@ Bot& RegionList::addRegion
...
+    dp = db_lookup(dbip, name, 0);
+
+    rt_db_get_internal(&region_internal, dp, dbip, bn_mat_identity, &rt_uniresource);
+    avs = &region_internal.idb_avs;
+    for (std::map<std::string, std::string>::const_iterator it = attributes.begin(); it != attributes.end(); it++)
+    {
+	bu_avs_add(avs, it->first.c_str(), it->second.c_str());
+    }
+
+    bu_avs_print(avs, name);
```
drossberg: Maybe for debugging purpose, but I wouldn't leave it in the final code.

src/conv/k-g/region_list.cpp
```diff
@@ -49,16 +49,53 @@ Bot& RegionList::addRegion
...
+
+    struct rt_db_internal     region_internal;
+    struct directory*         dp;
+    struct db_i*              dbip;
+    bu_attribute_value_set*   avs;
+
+    dbip = wdbp->dbip;
+    dp = db_lookup(dbip, name, 0);
+
+    rt_db_get_internal(&region_internal, dp, dbip, bn_mat_identity, &rt_uniresource);
```
drossberg: Why *bn_mat_identity* and not *NULL*?

src/conv/k-g/region_list.cpp
```diff
@@ -49,16 +49,53 @@ Bot& RegionList::addRegion
...
+}
+
+static void  writeAttributes
+(
+    rt_wdb* wdbp,
+    const char* name,
+    const std::map<std::string, std::string>& attributes
+) {
+
+    struct rt_db_internal     region_internal;
```
drossberg, line 67: Hmm, usually a `RT_DB_INTERNAL_INIT(&region_internal)` and `rt_db_free_internal(&region_internal)` is needed.

src/conv/k-g/region_list.cpp
```diff
@@ -49,16 +49,53 @@ Bot& RegionList::addRegion
...
+    struct directory*         dp;
+    struct db_i*              dbip;
+    bu_attribute_value_set*   avs;
```
drossberg: This is very much C-style. In C++, you can declare the variables where you need them: declaration = initialization

src/conv/k-g/region_list.h
```diff
@@ -32,24 +32,31 @@
...
+};
 
 class RegionList {
 public:
     RegionList(void);
 
     Bot& addRegion(const std::string& name);
 
-    Bot& getBot(const std::string& name);
+    Bot&                                getBot(const std::string& name);
```
drossberg: Why so many spaces?

src/conv/k-g/region_list.cpp
AliHaider93: actually I didn't check what it means, now that I checked we are not applying matrix transform on the object and the correct way is to put NULL.

src/conv/k-g/region_list.cpp
AliHaider93: I checked some other codes the RT_DB_INTERNAL_INIT wasn't used, but I did find that rt_db_free_internal(&region_internal) is used all the time. I will modify it

src/conv/k-g/region_list.cpp
AliHaider93: you are right. I will modify it

src/conv/k-g/region_list.cpp
AliHaider93: I think because the rt_db_internal is intialized using RT_DB_INTERNAL_INIT  inside the function rt_db_get_internal()

src/conv/k-g/region_list.h
AliHaider93: because there was the function const std::map<std::string, std::string> getAttributes() under it and I wanted to kep them next to each other. then we deleted getAttributes() and I forgot to bring it back :)

src/conv/k-g/region_list.cpp
AliHaider93: do think it's neccessary to check the Id returned by rt_db_get_internal()==ID_COMBINATION ?

src/conv/k-g/region_list.cpp
```diff
@@ -65,22 +65,19 @@ static void  writeAttributes
...
-    bu_attribute_value_set*   avs;
 
-    dbip = wdbp->dbip;
-    dp = db_lookup(dbip, name, 0);
+    struct db_i*      dbip = wdbp->dbip;
+    struct directory* dp = db_lookup(dbip, name, 0);
 
-    rt_db_get_internal(&region_internal, dp, dbip, bn_mat_identity, &rt_uniresource);
-    avs = &region_internal.idb_avs;
+    rt_db_get_internal(&region_internal, dp, dbip, NULL, &rt_uniresource);
```
AliHaider93: do you think that it's neccessary to check the id returend by 
rt_db_get_internal()==ID_COMBINATION?

drossberg: That's a good point. However, the ID is not so important, because *writeAttributes()* could write the attributes to any database object. But, you should check if the command was successful, i.e. the return value is >= 0.

Same for *dp*?

AliHaider93: they both need testing. I modified the code. please let me know if you have any notes

src/conv/k-g/region_list.cpp
```diff
@@ -49,16 +49,58 @@ Bot& RegionList::addRegion
...
+
+    struct db_i*      dbip = wdbp->dbip;
+    struct directory* dp = db_lookup(dbip, name, 0);
+
+    if (dp == RT_DIR_NULL) {
+	bu_log("writeAttributes() Cannot find %s\n", name);
+	return;
+    }
+
+    if (rt_db_get_internal(&region_internal, dp, dbip, NULL, &rt_uniresource) <= 0) {
```
drossberg: The documentation says that <0 means error, i.e. 0 would be okay too. In addition, even in case of error, *region_internal* has to be freed. Therefore, why not
```c++
if (rt_db_get_internal() >= 0) {
    // write the attributes
    // update database
}
else
    // error message

// free region_internal
```

AliHaider93: I did see in the code that most of the checks where rt_db_get_internal()<0, I think I should read more the documentaion is it the hacking guid you are talking about?. you are right the check you did is better I will modify.

AliHaider93: I can't update the database immediatly db5_update_attributes() should be called after having called bu_avs_add() which requires having the rt_db_internal already checked.

src/conv/k-g/region_list.cpp
```diff
@@ -49,16 +49,60 @@ Bot& RegionList::addRegion
...
+
+    struct db_i*      dbip = wdbp->dbip;
+    struct directory* dp = db_lookup(dbip, name, 0);
+
+    if (dp == RT_DIR_NULL) {
+	bu_log("writeAttributes() Cannot find %s\n", name);
+	return;
+    }
+
+    if (rt_db_get_internal(&region_internal, dp, dbip, NULL, &rt_uniresource) > 0) {
```
drossberg: **>= 0**?
BTW, the documentation I referred to where the comments in the header file include/rt/db_internal.h.

Hmm, it should be in the Doxygen documentation https://brlcad.org/docs/api/ (could be a bit outdated)
https://brlcad.org/docs/api/d2/de5/db__internal_8h.html#a5fabf39a9c6246df52172a92c2a2b221

This are the kind of references you can find only if you know that they exist :smiling_imp:

AliHaider93: I am really sorry it's >=0. I was distracted

PR #143, opened 2024-06-11
function to read node standard format

AliHaider93:
this function handels the standard node format I8,3E16.0,2F8.0


drossberg:
Almost perfect.
I think, you should put a comment above or inside `read_line_node_standard()` that says that the standard node line has the format I8,3E16.0,2F8.0 (8 digit integer, 3x 16 digit doubles, 2x 8 digit floats).

AliHaider93:
I did write the comment before then I deleted it :). I will modify and update the pull request

PR #144, opened 2024-06-14
draft realted to formats

AliHaider93:
in a key file there is mutliple places where the format might be determined. this is a first draft to handle multiple cases of different formats.

### DIFF
--- a/src/conv/k-g/k_parser.cpp
+++ b/src/conv/k-g/k_parser.cpp
@@ -194,6 +194,32 @@ static std::vector<std::string> read_line_node_i10
 }
 
 
+std::pair<std::string, std::string> split_key
+(
+    const char* key
+) {
+    std::pair<std::string, std::string> ret;
+    if (key != nullptr) {
+	std::string temp;
+	for (size_t i = 0; i < strlen(key); ++i) {
+	    if (key[i] == '=' ) {
+		if (temp.size() > 0) {
+		    ret.first = temp;
+		    temp.clear();
+		}
+	    }
+	    else if (key[i] != '=') {
+		temp += key[i];
+	    }
+	}
+	if (temp.size() > 0) {
+	    ret.second = temp;
+	}
+    }
+    return ret;
+}
+
+
 static std::vector<std::string> parse_line
 (
     const char* line
@@ -273,6 +299,7 @@ bool parse_k
 	std::string              line             = read_line(is);
 	ReadFormat               fileFormat       = ReadFormat::Standard;
 	ReadFormat               nodeFormat       = ReadFormat::Standard;
+	std::vector<std::string> nodeLines;//in old .k files a node with Long format is split in two lines.
 	std::vector<std::string> tokens;
 
 	if (line.size() > 0)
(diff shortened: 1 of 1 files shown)


src/conv/k-g/k_parser.cpp
```diff
@@ -254,6 +362,25 @@ bool parse_k
 			    else
 				std::cout << "Unexpected command " << tokens[0] << " in k-file " << fileName << "\n";
 			}
+			else if (command[0] == "KEYWORD") {
+			    if (tokens.size() == 2) {
```
drossberg: I looks like there could be many keys in any order?
Therefore I would recommend to loop through them with a for-loop, split the keys at the "=" character, and decide then about the format. The last one has it?

The split could be done in an own function: `std::pair<std::string, std::string> split_key(const char* key)`.

BTW, maybe some of the `const char*` could be replacved by `const std::string&`? :thinking:

src/conv/k-g/k_parser.cpp
```diff
@@ -267,7 +394,16 @@ bool parse_k
 			break;
 
 		    case KState::Node: {
-			tokens = read_line_node_standard(line.c_str());
+			if (nodeFormat == ReadFormat::Standard) {
+			    tokens = read_line_node_standard(line.c_str());
+			}
+			else if (nodeFormat == ReadFormat::Long) {
```
drossberg, line 504: You need to catch the two line format here. E.g.:
* have a variable for the first line
* if this variable is empty and the result of read_line_node_long() has the size 4 => write it to the variable and go on
* if this variable is not empty => add the result of read_line_node_long() to it, generate the node, and clear the variable

src/conv/k-g/k_parser.cpp
AliHaider93: yes there is usually mutiple keys (Formats), but I don't know if they come in any order. I will check it out.

src/conv/k-g/k_parser.cpp
AliHaider93: yes, thank you for the notes.

src/conv/k-g/k_parser.cpp
AliHaider93: it's actually more complicated, they *Keyword options are 1- Number i.e (*KEYWORD 120m) this requests for 120 million words of memeory. 2- memory2=j, i.e ( *KEYWORD 120m memory2=j) this defines the memory allocation in words for the first core and j defines the memory allocation in words for each of the remaining cores. 3- NCPU i.e (*KEYWORD NCPU=n) it defines the number Shared Memory Parallel (SMP) threads of "n" to be used for each processor during the analyhsis. all these dont affect the format. for the format you have the standard where is nothing written after the *KEYWORD. For Long format you find LONG=S or LONG=Y or LONG=K. For I10 there is written I10 then LONG=S i.e (*KEYWORD I10 LONG=S). I don't konw if splitting the keys with the function will be any more different. but I think it will be nicer. I will prepare it

src/conv/k-g/k_parser.cpp
```diff
@@ -72,12 +79,87 @@ static std::vector<std::string> read_line_node_standard
     const char* line
 ) {
     std::vector<std::string> ret;
+    bool commaSeparated = false;
+
+    if (line != nullptr) {
+	std::string temp;
+
+	for (size_t i = 0; i < strlen(line); ++i) {
+	    if (line[i] == ',' || line[i] == ' ') {
```
drossberg: Spaces can appear in the fixed format data too. Therefore, you cannot use it as a flag for comma-separated(?)

How about using `if (strchr(line, '\') != nullptr)` as criterion?

PR #146, opened 2024-07-04
Case insensitive keywords

AliHaider93:
The keywords in .k file are case insensitive.

### DIFF
--- a/src/conv/k-g/k_parser.cpp
+++ b/src/conv/k-g/k_parser.cpp
@@ -343,8 +343,10 @@ static std::vector<std::string> parse_command
 		    temp.clear();
 		}
 	    }
-	    else
-		temp += command[i];
+	    else {
+		char uper = toupper(command[i]);
+		temp += uper;
+	    }
 	}
 
 	if (temp.size() > 0)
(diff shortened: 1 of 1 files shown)


drossberg:
Would be "Node" be a valid keyword too? In this case, using `toupper()` would be better.

drossberg:
Or something similar, see e.g. the Notes in https://en.cppreference.com/w/cpp/string/byte/toupper.

AliHaider93:
> Would be "Node" be a valid keyword too? In this case, using `toupper()` would be better.

your guess is right :), Node Also works, also NOde works, I will use toupper().

AliHaider93:
I tried using special characters for the part name like 
image
image
and LS-PrePost doesnt care about them, so I think to upper is sfficent with a condition to change the parts names into a sequence of numbers if toupper()  incounters special charachters.

AliHaider93:
it's unnessessary to check for weird Part name, mged doesn't care about it 
image

drossberg:
`toupper()` is applied to the first token of lines starting with a '*' only, right? Therefore, other stuff, like the part names, aren't touched.

drossberg:
Very nice patch. In fact, it could  be a one-liner too.

PR #147, opened 2024-07-05
Geometry class

AliHaider93:
A class to incapsulate the geometry of LS-DYNA part.

### DIFF
--- a/src/conv/k-g/arbs.cpp
+++ b/src/conv/k-g/arbs.cpp
@@ -89,6 +89,7 @@ std::vector<std::string> Arbs::write
 
     for (std::map<std::string, rt_arb_internal>::iterator it = arbs.begin(); it != arbs.end(); it++) {
 	std::string arbName = name;
+	arbName += ".";
 	arbName += it->first;
 	arbName += ".arb";
 	ret.push_back(arbName);
--- a/src/conv/k-g/arbs.cpp
+++ b/src/conv/k-g/arbs.cpp
@@ -51,14 +51,14 @@ void Arbs::addArb(
     rt_arb_internal temp;
     temp.magic = RT_ARB_INTERNAL_MAGIC;
 
-    VSET(temp.pt[0], point1[0], point1[1], point1[2]);
-    VSET(temp.pt[1], point2[0], point2[1], point2[2]);
-    VSET(temp.pt[2], point3[0], point3[1], point3[2]);
-    VSET(temp.pt[3], point4[0], point4[1], point4[2]);
-    VSET(temp.pt[4], point5[0], point5[1], point5[2]);
-    VSET(temp.pt[5], point6[0], point6[1], point6[2]);
-    VSET(temp.pt[6], point7[0], point7[1], point7[2]);
-    VSET(temp.pt[7], point8[0], point8[1], point8[2]);
+    VMOVE(temp.pt[0], point1);
+    VMOVE(temp.pt[1], point2);
+    VMOVE(temp.pt[2], point3);
+    VMOVE(temp.pt[3], point4);
+    VMOVE(temp.pt[4], point5);
+    VMOVE(temp.pt[5], point6);
+    VMOVE(temp.pt[6], point7);
+    VMOVE(temp.pt[7], point8);
 
     arbs[arbName] = temp;
 }
--- a/src/conv/k-g/arbs.cpp
+++ b/src/conv/k-g/arbs.cpp
@@ -1,0 +1,107 @@
+/*                         A R B S . C P P
+ * BRL-CAD
+ *
+ * Copyright (c) 2024 United States Government as represented by
+ * the U.S. Army Research Laboratory.
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public License
+ * version 2.1 as published by the Free Software Foundation.
+ *
+ * This library is distributed in the hope that it will be useful, but
+ * WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this file; see the file named COPYING for more
+ * information.
+ */
+/** @file arbs.cpp
+ *
+ * LS Dyna keyword file to BRL-CAD converter:
+ * intermediate arbs data implementation
+ */
+
+#include "common.h"
+
+#include "arbs.h"
+
+
+Arbs::Arbs(void):name() {
+    //arb_internal.magic = RT_ARBN_INTERNAL_MAGIC;
+    //arb_internal.magic = RT_ARB_INTERNAL_MAGIC;
+}
+
+Arbs::~Arbs(void)
+{
+    /*if (arb_internal.pt != nullptr)
+	bu_free(arb_internal.pt, "Solid::~Solid: pt");*/
+}
+
+void Arbs::setName(const char* value)
+{
+    name = value;
+}
+
+void Arbs::addArb(
+    const char* arbName,
+    const point_t& point1,
+    const point_t& point2,
+    const point_t& point3,
+    const point_t& point4,
+    const point_t& point5,
+    const point_t& point6,
+    const point_t& point7,
+    const point_t& point8
+) {
+    rt_arb_internal temp;
+    temp.magic = RT_ARB_INTERNAL_MAGIC;
+
+    VSET(temp.pt[0], point1[0], point1[1], point1[2]);
+    VSET(temp.pt[1], point2[0], point2[1], point2[2]);
+    VSET(temp.pt[2], point3[0], point3[1], point3[2]);
+    VSET(temp.pt[3], point4[0], point4[1], point4[2]);
+    VSET(temp.pt[4], point5[0], point5[1], point5[2]);
+    VSET(temp.pt[5], point6[0], point6[1], point6[2]);
+    VSET(temp.pt[6], point7[0], point7[1], point7[2]);
+    VSET(temp.pt[7], point8[0], point8[1], point8[2]);
+
+    arbs[arbName] = temp;
+}
+
+const char* Arbs::getName(void) const
+{
+    return name.c_str();
+}
+
+std::map<std::string, rt_arb_internal> Arbs::getArbs(void) const
+{
+    return arbs;
+}
+
+
+void Arbs::write
+(
+    rt_wdb* wdbp
+){
+    for (std::map<std::string, rt_arb_internal>::iterator it = arbs.begin(); it != arbs.end(); it++) {
+	rt_arb_internal* arb_wdb;
+
+	BU_GET(arb_wdb, rt_arb_internal);
+	arb_wdb->magic = RT_ARB_INTERNAL_MAGIC;
+	*arb_wdb = it->second;
+
+	wdb_export(wdbp, (it->first).c_str(), arb_wdb, ID_ARB8, 1.);
+    }
+}
+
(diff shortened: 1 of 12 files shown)


src/conv/k-g/k_parser.h
```diff
@@ -25,12 +25,14 @@
 
 #ifndef KPARSER_INCLUDED
 #define KPARSER_INCLUDED
+#define FIRST_NODE 2
```
drossberg: ? Where is it used?

drossberg: You see that you are getting into trouble with the primitives names. How about not having a `setName()` but `setBaseName()` and not `getName()` but `getNames()` which returns a list of names? In case of the Bot, it would be the base name with a ".bot" attached. In case of the Arbs, it could be a list of base-name + "." + number + ".arb".

AliHaider93: > ? Where is it used?

it's not used any more, I was trying to make a loop and I defined it. I will rimove it. Edit: it is used in k_parser.cpp to fill the nodes in the vector using a loop 

image

I can just use i_n+2 because it's a fixed format, but this seamed nicer. what do you think?

AliHaider93: > You see that you are getting into trouble with the primitives names. How about not having a `setName()` but `setBaseName()` and not `getName()` but `getNames()` which returns a list of names? In case of the Bot, it would be the base name with a ".bot" attached. In case of the Arbs, it could be a list of base-name + "." + number + ".arb".

okay, this seems to be more coherent, I am working on it.

AliHaider93: the only problem with setBaseName() is in the case of arbs the elements numbers doesn't have to be consequent starting from (1 to n). so if we want to preserve the number of Hex element we have to keep naming the arbs following the .k file. I will update the pull request and then you can tell me what you think

AliHaider93: I updated with getNames() and setBaseName(), I will change also Arbs::write() and Bot::write().

drossberg: > > ? Where is it used?
> 
> it's not used any more, I was trying to make a loop and I defined it. I will remove it. Edit: it is used in k_parser.cpp to fill the nodes in the vector using a loop
> 
> image
> 
> I can just use i_n+2 because it's a fixed format, but this seamed nicer. what do you think?

Okay, that's a reason. However, in this case, it should be kept in k_parser.cpp (declaration/definition) and a `const size_t`, e.g.
```c++
const size_t FirstNode = 2;
```

src/conv/k-g/arbs.cpp
```diff
@@ -81,19 +81,28 @@ std::map<std::string, rt_arb_internal> Arbs::getArbs(void) const
...
-void Arbs::write
+std::vector<std::string> Arbs::write
 (
     rt_wdb* wdbp
 ){
+    std::vector<std::string> ret;
+
     for (std::map<std::string, rt_arb_internal>::iterator it = arbs.begin(); it != arbs.end(); it++) {
+	std::string arbName = name;
+	arbName += it->first;
```
drossberg, line 88: Maybe, putting a point "." between the base name and the index?

AliHaider93: yes it's better , I will update

AliHaider93: I just have a question about the constructors and destructors, first the arb constructor 
image
is RT_ARBN_INTERNAL_MAGIC the one to go even with arb4, arb5, arb6, arb7? 
and for Geomtry I  think it doesn't need a constructor or a destructor or am I wrong?

AliHaider93: then I will handle the other arbs and I can clean the code

drossberg: You need the RT_ARB_INTERNAL_MAGIC here. This is for the rt_arb_internal.

arbn is very different from the arb4 ... arb8.

drossberg: For the Geometry class, the standard constructor and destructor should do.

src/conv/k-g/k-g.cpp
```diff
@@ -182,6 +181,228 @@ int main
...
+				point7[Y] = kData.nodes[n7].y * factor;
+				point7[Z] = kData.nodes[n7].z * factor;
+
+				point8[X] = kData.nodes[n8].x * factor;
+				point8[Y] = kData.nodes[n8].y * factor;
+				point8[Z] = kData.nodes[n8].z * factor;
+
+				std::string arbNumber = std::to_string(*itr);
+
+				geometry.addArb(arbNumber.c_str(), point1, point2, point3, point4, point5, point6, point7, point8);
```
drossberg: That's a lot of duplicated code. How about writing a helper function `AddArb()`, which gets n1, ..., n8 as parameters?

src/conv/k-g/arbs.cpp
```diff
@@ -0,0 +1,112 @@
...
+    temp.magic = RT_ARB_INTERNAL_MAGIC;
+
+    VSET(temp.pt[0], point1[0], point1[1], point1[2]);
+    VSET(temp.pt[1], point2[0], point2[1], point2[2]);
+    VSET(temp.pt[2], point3[0], point3[1], point3[2]);
+    VSET(temp.pt[3], point4[0], point4[1], point4[2]);
+    VSET(temp.pt[4], point5[0], point5[1], point5[2]);
+    VSET(temp.pt[5], point6[0], point6[1], point6[2]);
+    VSET(temp.pt[6], point7[0], point7[1], point7[2]);
+    VSET(temp.pt[7], point8[0], point8[1], point8[2]);
```
drossberg: Would it be possible, to use VMOVE() here?

src/conv/k-g/k-g.cpp
AliHaider93: you are right, working on it.

AliHaider93: I also relized that I made a mistake in the condition for arb4, when the nodes.size()==4 it is also the shell element that we devided into two triangles

src/conv/k-g/arbs.cpp
AliHaider93: yes it's possible, I updated the code

AliHaider93: I also wrote the function AddArb, let me know what you think about it. for arb4 I will resolve it later, I am trying to undrstand the logic in LS-DYNA Solid elment.

src/conv/k-g/k-g.cpp
```diff
@@ -82,45 +130,109 @@ int main
...
+				}
+				else {
+				    geometry.addTriangle(point1, point2, point3);
+				    geometry.addTriangle(point1, point3, point4);
+				}
 			    }
-			    else {
-				bot.addTriangle(point1, point2, point3);
-				bot.addTriangle(point1, point3, point4);
+			    else if (kData.elements[*itr].nodes.size() == 8) {
```
drossberg, line 175: Is here an issue with a closing bracket?

src/conv/k-g/k-g.cpp
```diff
@@ -82,45 +130,109 @@ int main
...
+				    geometry.addTriangle(point1, point2, point3);
+				}
+				else {
+				    geometry.addTriangle(point1, point2, point3);
+				    geometry.addTriangle(point1, point3, point4);
+				}
 			    }
-			    else {
-				bot.addTriangle(point1, point2, point3);
-				bot.addTriangle(point1, point3, point4);
```
drossberg, line 122: Is here an issue with a closing bracket?

AliHaider93: No, I don't think so, I have tested the code with a solid element and shell element. I will retest now to see if there is something wrong

AliHaider93: I just tested it, no problems.

src/conv/k-g/region_list.cpp
```diff
@@ -99,26 +97,30 @@ void RegionList::create
...
+	    mk_addmember(names[i].c_str(), &(geometry_head.l), NULL, WMOP_UNION);
+	    mk_addmember(names[i].c_str(), &(all_head.l), NULL, WMOP_UNION);
+	}
+	mk_lfcomb(wdbp, geometry.getBaseName(), &geometry_head, 0);
 
 	if (region_attributes.size() > 0) {
 	    writeAttributes(wdbp, region_name.c_str(), region_attributes);
 	}
 
 	mk_addmember(region_name.c_str(), &(all_head.l), NULL, WMOP_UNION);
```
drossberg: There is a tab in this line, which should not be there.

src/conv/k-g/arbs.h
```diff
@@ -0,0 +1,76 @@
...
+
+
+class Arbs {
+public:
+    Arbs(void);
+
+    void        setName(const char* value);
+
+    void        addArb(
+			const char*    arbName,
```
drossberg: Is this line break intentionally? (similar in arbs.cpp)

src/conv/k-g/region_list.cpp
```diff
@@ -99,26 +97,30 @@ void RegionList::create
...
+	std::vector<std::string> names = geometry.write(wdbp);
 
-	wmember bot_head;
-	BU_LIST_INIT(&bot_head.l);
+	wmember geometry_head;
+	BU_LIST_INIT(&geometry_head.l);
 
-	mk_addmember(bot.getName(), &(bot_head.l), NULL, WMOP_UNION);
-	mk_lfcomb(wdbp, region_name.c_str(), &bot_head, 0);
+	for (int i = 0; i < names.size(); i++) {
```
drossberg: I'm getting a compilation error on Debian bookworm here:
```
/home/rossberg/Devel/BRL-CAD/brlcad.devel_k-g/src/conv/k-g/region_list.cpp: In member function ‘void RegionList::create(rt_wdb*)’:
/home/rossberg/Devel/BRL-CAD/brlcad.devel_k-g/src/conv/k-g/region_list.cpp:111:27: error: comparison of integer expressions of different signedness: ‘int’ and ‘std::vector<std::__cxx11::basic_string<char> >::size_type’ {aka ‘long unsigned int’} [-Werror=sign-compare]
  111 |         for (int i = 0; i < names.size(); i++) {
      |                         ~~^~~~~~~~~~~~~~
```
Changing `int` to `size_t` solves this.

src/conv/k-g/region_list.cpp
```diff
@@ -140,7 +142,7 @@ Bot& RegionList::getBot
 (
     const std::string& name
 ) {
-    return m_list[name].bot;
+    return m_list[name].geometry.getBot();
```
drossberg, line 145: I'm getting a compilation error on Debian bookworm here:
```
/home/rossberg/Devel/BRL-CAD/brlcad.devel_k-g/src/conv/k-g/region_list.cpp: In member function ‘Bot& RegionList::getBot(const std::string&)’:
/home/rossberg/Devel/BRL-CAD/brlcad.devel_k-g/src/conv/k-g/region_list.cpp:145:40: error: cannot bind non-const lvalue reference of type ‘Bot&’ to an rvalue of type ‘Bot’
  145 |     return m_list[name].geometry.getBot();
      |            ~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~
```
`getBot()` doesn't return a reference.

src/conv/k-g/geometry.cpp
```diff
@@ -0,0 +1,109 @@
...
+    m_arbs.addArb(arbName, point1, point2, point3, point4, point5, point6, point7, point8);
+}
+
+
+const char* Geometry::getBaseName(void) const{
+    return name.c_str();
+}
+
+
+Bot Geometry::getBot(void) const{
```
drossberg: You should change the return type to reference here.

PR #151, opened 2024-07-27
distinction between command and option

AliHaider93:
this is a way to make a distenction between a command and it's options. and eventually this is a necessary code to handle all kinds of elements in LS-DYNA


src/conv/k-g/k_parser.cpp
```diff
@@ -258,17 +286,51 @@ bool parse_k
...
+				    //elementOptions.insert(elementOptions.end(), command.begin() + 3, command.end());//ELEMENT_BEAM_SOURCE doesn't have options 
 				}
 				else {
 				    state = KState::Element_Beam;
 
 				    elementOptions.insert(elementOptions.end(), command.begin() + 2, command.end());
+
+				    if (elementOptions.size() > 0) {
+					for (size_t i_o = 0; i_o < elementOptions.size(); ++i_o) {
+					    if (elementOptions[i_o] == "ThICKNESS") {
```
drossberg: Shouldn't it be a capital H in ThICKNESS?

src/conv/k-g/k_parser.cpp
```diff
@@ -477,6 +759,89 @@ bool parse_k
...
+				element.nodes.push_back(stoi(tokens[i_n + FirstNode]));
+			    }
+
+			    data.elements[eid] = element;
+
+			    int pid = stoi(tokens[1]);
+			    data.parts[pid].elements.insert(eid);
+			    break;
+			}
+			else if ((elementBeamOptions.size()) > 0 && (elementOptionsCounter < elementBeamOptions.size())) {
```
drossberg: Hmm, if `elementBeamOptions.size()` is greater than `elementOptionsCounter`, then it must be greater than 0 too.

src/conv/k-g/k_parser.cpp
```diff
@@ -477,6 +759,89 @@ bool parse_k
...
+			    Element_Beam_Options currentOption = elementBeamOptions[elementOptionsCounter];
+			    switch (currentOption)
+			    {
+			    case Element_Beam_Options::Thickness: {
+				//handle the thickness line
+				elementOptionsCounter++;
```
drossberg: I prefer `++object` whenever possible:
- `++object` returns the value after `object` was incremented, i.e. a reference to `object`
- `object++` returns the value of `object` before it was incremented; to do this, a copy of `object` has to be created to hold this value, which can be expensive
This doesn't matter in your case here case, but it isn't always obvious, if the extra costs of `object++` are high.

src/conv/k-g/k_parser.cpp
AliHaider93: yes, I was expecting some mistakes, I will check everything better if you think this is the right way to go

src/conv/k-g/k_parser.cpp
AliHaider93: you are right, I will fix it

src/conv/k-g/k_parser.cpp
AliHaider93: this one I didn't know about. thank you for the advice :)

src/conv/k-g/k_parser.cpp
```diff
@@ -477,6 +758,329 @@ bool parse_k
...
+
+			    break;
+			}
+			case 4: {
+			    if (tokens.size() < 4) {
+				std::cout << "Too short ELEMENT_BEARING in k-file " << fileName << "\n";
+				break;
+			    }
+
+			    bearing.ineerGroveRadiusToBallDiameterRatioOrRollerLength = stof(tokens[1]);
```
drossberg: Typing error: ineer -> inner?

src/conv/k-g/geometry.cpp
```diff
@@ -95,15 +95,31 @@ Arbs& Geometry::getArbs(void) {
...
+
+
+// Local Variables:
+// tab-width: 8
+// mode: C++
+// c-basic-offset: 4
+// indent-tabs-mode: t
+// c-file-style: "stroustrup"
+// End:
+// ex: shiftwidth=4 tabstop=8
```
drossberg: Source files should end with a new-line.

src/conv/k-g/pipe.cpp
```diff
@@ -0,0 +1,105 @@
...
+
+
+// Local Variables:
+// tab-width: 8
+// mode: C++
+// c-basic-offset: 4
+// indent-tabs-mode: t
+// c-file-style: "stroustrup"
+// End:
+// ex: shiftwidth=4 tabstop=8
```
drossberg: ditto

src/conv/k-g/pipe.h
```diff
@@ -0,0 +1,65 @@
...
+
+
+// Local Variables:
+// tab-width: 8
+// mode: C++
+// c-basic-offset: 4
+// indent-tabs-mode: t
+// c-file-style: "stroustrup"
+// End:
+// ex: shiftwidth=4 tabstop=8
```
drossberg: ditto

src/conv/k-g/k-g.cpp
```diff
@@ -136,7 +137,38 @@ int main
...
+				    KSectionBeam beamSection = kData.sectionsBeam[section];
+
+				    if (beamSection.CST == 1) {
+					pipePoint point1;
+					pipePoint point2;
+
+
+					int n1 = kData.elements[*itr].nodes[0];
+					int n2 = kData.elements[*itr].nodes[1];
+					int n3 = kData.elements[*itr].nodes[2];
```
drossberg: The **n3** variable is unused. I get a compilation error here.

PR #153, opened 2024-08-05
region color

AliHaider93:
implementation of random coloring for the regions

### DIFF
--- a/src/conv/k-g/region_list.cpp
+++ b/src/conv/k-g/region_list.cpp
@@ -120,8 +120,7 @@ void RegionList::create
 	    mk_addmember(names[i].c_str(), &(all_head.l), NULL, WMOP_UNION);
 	}
 
-	mk_lcomb(wdbp, geometry.getBaseName(), &geometry_head, 1, NULL, NULL, rgb, region_id);
-	++region_id;
+	mk_lcomb(wdbp, geometry.getBaseName(), &geometry_head, 1, NULL, NULL, rgb, ++region_id);
 
 	if (region_attributes.size() > 0) {
 	    writeAttributes(wdbp, region_name.c_str(), region_attributes);
--- a/src/conv/k-g/region_list.cpp
+++ b/src/conv/k-g/region_list.cpp
@@ -96,6 +96,8 @@ void RegionList::create
     wmember all_head;
     BU_LIST_INIT(&all_head.l);
 
+    int region_id = 0;
+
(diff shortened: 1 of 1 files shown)


src/conv/k-g/region_list.cpp
```diff
@@ -102,17 +102,23 @@ void RegionList::create
...
+	unsigned char rgb[3];
+	bu_color_to_rgb_chars(&regionColor, rgb);
+
 	for (size_t i = 0; i < names.size(); i++) {
 	    mk_addmember(names[i].c_str(), &(geometry_head.l), NULL, WMOP_UNION);
 	    mk_addmember(names[i].c_str(), &(all_head.l), NULL, WMOP_UNION);
 	}
-	mk_lfcomb(wdbp, geometry.getBaseName(), &geometry_head, 0);
+
+	mk_comb(wdbp, geometry.getBaseName(), &geometry_head.l, 1, (char*)NULL, (char*)NULL, rgb, ID_COMBINATION, 0, 0, 0, 0, 0, 0);
```
drossberg: * Is this `(char*)` cast really necessary?
* The **id** or **region_id** is usually a unique number of a region, and not related to ID_COMBINATION. The "default" would be an increasing integer.
* Why not `mk_lcomb()`?

AliHaider93: checking and will update.

src/conv/k-g/region_list.cpp
```diff
@@ -96,23 +96,32 @@ void RegionList::create
...
+	bu_color_to_rgb_chars(&regionColor, rgb);
+
 	for (size_t i = 0; i < names.size(); i++) {
 	    mk_addmember(names[i].c_str(), &(geometry_head.l), NULL, WMOP_UNION);
 	    mk_addmember(names[i].c_str(), &(all_head.l), NULL, WMOP_UNION);
 	}
-	mk_lfcomb(wdbp, geometry.getBaseName(), &geometry_head, 0);
+
+	mk_lcomb(wdbp, geometry.getBaseName(), &geometry_head, 1, NULL, NULL, rgb, region_id);
+	++region_id;
```
drossberg: One more (new) thing :smiling_imp:: Please, start the region IDs with 1. You can do this by writing ++region_id inside the mk_lcomb: `mk_lcomb(wdbp, geometry.getBaseName(), &geometry_head, 1, NULL, NULL, rgb, ++region_id);`

AliHaider93: sure, will update.

PR #164, opened 2024-10-21
sync with main branch

drossberg:
need the BRLCAD_EXT_DIR patch here


CMakeLists.txt
brlcad, line 1: Thank you!  I have hit this repeatedly and didn't really catch what was going on wrong.

PR #170, opened 2025-01-13
Update README.Linux

Biswaranjan-Dash:
Cmake requirement for ubuntu/debian changed to 3.25.0 as the build has a new signature style which fails if build with a cmake version lower than 3.25.0

### DIFF
--- a/doc/README.Linux
+++ b/doc/README.Linux
@@ -67,7 +67,7 @@ Following the build instructions in the INSTALL file.  You will need:
 gcc (6+, e.g. 6.0.3)
 g++ (6+, e.g. 6.0.3)
 make (e.g. gnu make 3.8.0)
-cmake (3.0.2 or newer)
+cmake (3.25.0 or newer)
 
 All three of those have implicit dependencies on other packages.
 


starseeker:
Um.  What was the error?  I was under the impression we were viable with 3.18, but maybe something crept in...

Biswaranjan-Dash:
yeah, i think there was some change in how the directories were accessed. With cmake 3.22 it was continously producing  directory doesn't exist and trycompile failed. On upgrading the cmake version to 3.31, produced no error. And going though the cmake documentation, 3.25 version had some changes in the command signature style. Maybe this caused the issue. So, mentioning cmake 3.25 would be a better option for developers. 
image

starseeker:
Ah, I see it now.  I adjusted the src/libged/CMakeLists.txt file to use a different try_compile signature: https://github.com/BRL-CAD/brlcad/commit/e9d8e5a551d4851108a3d4417c6bd945971f2900

starseeker:
Looks like the actual minimum now is 3.20 - I'll update our logic and docs accordingly.

starseeker:
@Biswaranjan-Dash Let me know if https://github.com/BRL-CAD/brlcad/commit/1734c58ba5dc04d31216152baed782d100f7adc8 doesn't cover it - I gave it a whirl here with https://cmake.org/files/v3.20/cmake-3.20.6-linux-x86_64.tar.gz and it seemed to configure successfully.

PR #181, opened 2025-04-11
drafted an algorithm for plate mode bot tessillation

drossberg:
there is no perfect solution for plate mode bots, because the ray-trace entry and exit points may vary depending on the ray direction; there is no well defined boundary representation
however, let's do at least something


starseeker:
@drossberg How does this compare to what we're doing with the bot extrude subcommand in src/libged/bot/extrude.cpp (which uses src/librt/primitives/bot/plate.cpp) ?

starseeker:
I'm in the process of testing some improvements to the performance of that code as well using Manifold v3:  https://github.com/BRL-CAD/brlcad/tree/manifold3

starseeker:
Algorithmically plate.cpp is a brute force solution, so if you've got something more elegant I'm definitely interested, but just need to know the pros and cons since plate.cpp is at least able to handle something like the FAA's GenericTwin example reasonably well (if a bit slowly).

drossberg:
If I understood it correctly, plate.cpp replaces each triangle by something similar to an arb6, followed by a Boolean unification. My solution tries to find a "mean" normal for each vertex, where it constructs the tessellation from, using these normals in + ind - directions. I did it this way because the Boolean BREP algorithms in libnmg are fairly bad.

However, I've found already a first error in my algorithm. And it seems to have room for improvements. But the advantage would be still to avoid the Boolean operations, at least here.

starseeker:
We're not using NMG booleans for plate.cpp - we're using the Manifold library, which has much better success rates:  https://github.com/elalish/manifold

PR #185, opened 2025-04-26
Fix Unsafe Floating-Point Comparison in bn_nextafterf_up and bn_nexta…

### LINKED ISSUE #184: -Werror=float-equal  is triggered  when using !val to complare floting point values
The use of the `-Werror=float-equal` flag, which causes all warnings related to floating-point comparisons `(==` or ` !=`) to be treated as errors. In C and C++ programming, comparing floating-point values directly using equality (e.g., `==` or` !=`) is generally not safe due to the precision issues inherent in floating-point representation.

Looking at the` ulp.c` file, we can see that it is comparing floating-point values with `== `and `!=` directly, which is causing the warnings and errors related to unsafe floating-point comparisons.

Image

Forchapeatl:
This PR addresses an issue where the code in `libbn/ulp.c` used unsafe comparisons for floating-point values, specifically in the `bn_nextafterf_up` and `bn_nextafterf_dn` functions. The use of `!val` to check for zero is problematic, as it assumes that `val` is exactly zero, which can cause inaccuracies due to floating-point precision issues.

Fixes #184 

- Replaced `!val` checks with `fabs(val) < FLT_EPSILON` in both the `bn_nextafterf_up` and `bn_nextafterf_dn` functions. This ensures that comparisons for near-zero floating-point values are done safely using a tolerance (`FLT_EPSILON`).
- This change prevents the unsafe comparison and ensures compatibility with floating-point arithmetic, addressing both the precision issues and the compiler warning.

Screenshot from 2025-04-26 05-23-32

### DIFF
--- a/src/libbn/ulp.c
+++ b/src/libbn/ulp.c
@@ -245,7 +245,7 @@ bn_nextafter_up(double val)
 	return val;
 
     /* unify both +0 and -0 into +0’s next subnormal */
-    if (!val) {
+    if (fabs(val) < DBL_EPSILON) {
         bits.u = 1ULL; /* raw 0x0000…001 */
         return bits.d;
     }
@@ -265,7 +265,7 @@ bn_nextafter_dn(double val)
 	return val;
 
     /* unify both +0 and -0 into -0’s next subnormal */
-    if (!val) {
+    if (fabs(val) < DBL_EPSILON) {
         /* raw, largest negative subnormal */
         bits.i = (int64_t)0x8000000000000001ULL;
         return bits.d;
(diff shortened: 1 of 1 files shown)


Forchapeatl:
@drossberg   and @brlcad  please have a look at my PR

drossberg:
Hi @Forchapeatl, thanks for caring. This issue was however "fixed" with caddf075af462da6b523daf40c8249aa64a3ea59.

The point is that checking doubles for equality is often wrong, but not always. Especially the value 0. is often used as a flag, e.g. for uninitialized values. (But I'm not sure, if -0. == +0., as the comments suggest.)

Comparing with ~_EPSILON is definitely wrong here. The values must be 0.. ~_MIN would probable do the job (and catch +/- 0. anyway). Moreover, some magic numbers could possibly be replaced by ~_MIN and ~_MAX.

brlcad:
@drossberg I had to look that one up myself and indeed +0 == -0 

@Forchapeatl thank you for trying to fix a current build issue.  The effort is appreciated even if the suggestion is ultimately OBE (and incorrect).  Encourage you to find another area of work that aligns with your experience or that you can develop into experience as that's a great way to get involved in any open source community.

PR #187, opened 2025-04-29
Fixed Unkown PLot error

### LINKED ISSUE #183: Unknown error when creating plots
When I try to create plots I get error but the file creates **successfully**.  BUT says **Error** 

###  Where does the error come from  ?  I would like to debug further 

-  **Error logs image** 

Image

- **Successful file creation** 

Image

Forchapeatl:
Fixes #183 

### Changes:
- **`dl_plot` function**: 
  - Returns `0` on success and `1` on error.
  - Introduces error checks for invalid pointers and operations ( null checks, invalid arguments / vertex data).
  
- **`ged_plot_core` function**:
  - Checks the return value from `dl_plot` and handles error states by returning `BRLCAD_ERROR` if an error occurs during plotting.

image

### DIFF
--- a/src/libged/plot/plot.c
+++ b/src/libged/plot/plot.c
@@ -43,7 +43,7 @@ extern FILE *popen(const char *command, const char *type);
 extern int pclose(FILE *stream);
 #endif
 
-int
+void
 dl_plot(struct bu_list *hdlp, FILE *fp, mat_t model2view, int floating, mat_t center, fastf_t scale, int Three_D, int Z_clip)
 {
     struct display_list *gdlp;
@@ -55,10 +55,6 @@ dl_plot(struct bu_list *hdlp, FILE *fp, mat_t model2view, int floating, mat_t ce
     static vect_t fin;
     static vect_t start;
     int Dashing;                        /* linetype is dashed */
-
-    if (!hdlp || !fp || !model2view) {
-        return 1;       /* Error: Invalid arguments */
-    }
 
     if (floating) {
         pd_3space(fp,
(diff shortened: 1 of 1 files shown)


Forchapeatl:
@brlcad  and @drossberg  please  have  look at my PR

drossberg:
Hi @Forchapeatl, which issue do you want to solve with your patch? Which of the checks wasn't already done in ged_plot_core()? You checks look redundant to me.

Forchapeatl:
Good day Sir, @drossberg ,
Before I tried to create plots , the plot  file gets created successfully, BUT says `Error.`

image

This happened  because `ged_plot_core` was designed to return an ERROR

https://github.com/BRL-CAD/brlcad/blob/2f1393c6fb268c29a652ec293c39a5fd93033e53/src/libged/plot/plot.c#L292

This PR  introduces vulnerability   checks on `dl_plot`  . The `ged_plot_core` now returns  `BRLCAD_ERROR` or `BRLCAD_OK` depending on the state of  `dl_plot` .

drossberg:
Ah, I see. Well done, you found a bug: There is no way to return success in ged_plot_core().

Unfortunately, you do a bit too much in your patch. E.g., you check the file pointer **fp** for NULL in dl_plot(). But this isn't necessary. It will never be NULL, because **fp** will always be checked right after creation in ged_plot_core(). Similar is true for **model2view**. GED_CHECK_VIEW() checks it implicitly. That's why I asked you, which of the checks wasn't already done in ged_plot_core()? I could imagine that all pointers are guaranteed to be not NULL.

I suspect that changing the last BRLCAD_ERROR to BRLCAD_OK would fix the bug.

Forchapeatl:
Sir, please I have made the requested changes.

drossberg:
This should do it.

I hope that you tracked down the test for **hdlp** not being NULL. Being able to read and understand existing code is very important if you want to work as a software engineer. I recommend to exercise this, e.g. whenever you work with open source software, especially if you want to contribute to it.

PR #189, opened 2025-06-06
Seatbelt first implementation

BraghinAndrea:
It can successfully convert Seatbelt elements from .k files. 
I believe k-g.cpp main loop should host the seatbelt loop as well but I had a tough time figuring out the right way to do so. Any feedback will be precious!

Converted geometry follows LS-PrePost reference coordinate system, tuning of scalar factor might be needed in the near future.

### DIFF
--- a/src/conv/k-g/k-g.cpp
+++ b/src/conv/k-g/k-g.cpp
@@ -39,17 +39,18 @@
 
 static void AddArb
 (
-    int         n1,
-    int         n2,
-    int         n3,
-    int         n4,
-    int         n5,
-    int         n6,
-    int         n7,
-    int         n8,
-    KData&      kData,
-    std::string arbNumber,
-    Geometry&   geometry
+    int          n1,
+    int          n2,
+    int          n3,
+    int          n4,
+    int          n5,
+    int          n6,
+    int          n7,
+    int          n8,
+    KData&       kData,
+    std::string  arbNumber,
+    Geometry&    geometry,
+	const double factor
 ) {
     point_t point1;
     point_t point2;
@@ -59,8 +60,6 @@ static void AddArb
     point_t point6;
     point_t point7;
     point_t point8;
-
-    const double factor = 10.0;
 
     point1[X] = kData.nodes[n1].x * factor;
     point1[Y] = kData.nodes[n1].y * factor;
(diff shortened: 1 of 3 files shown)


src/conv/k-g/k-g.cpp
```diff
@@ -39,17 +39,18 @@
...
+    int          n3,
+    int          n4,
+    int          n5,
+    int          n6,
+    int          n7,
+    int          n8,
+    KData&       kData,
+    std::string  arbNumber,
+    Geometry&    geometry,
+	const double factor
```
AliHaider93, line 53: Here it is actually the correct way to pass the factor to the function. but the addarb function is not related to the seatbelt PR. non the less it's a small modification and it formalize the addarb and addseatbelt functions. Maybe it might be possible to use 1 function for both cases.

src/conv/k-g/k-g.cpp
```diff
@@ -131,58 +201,74 @@ int main
 			std::string partName = std::to_string(it->first) + "_" + (it->second).title;
 
 			Geometry& geometry = regions.addRegion(partName);
-			int    section = (it->second).section;
+			int    sectionid = (it->second).section;
```
AliHaider93, line 204: Not related to PR topic

src/conv/k-g/k-g.cpp
```diff
@@ -131,58 +201,74 @@ int main
...
 
-			if (section > 0) {
-			    double thick1 = kData.sections[section].thickness1;
-			    double thick2 = kData.sections[section].thickness1;
-			    double thick3 = kData.sections[section].thickness1;
-			    double thick4 = kData.sections[section].thickness1;
+			std::cout << sectionid << std::endl;
```
AliHaider93, line 206: Here it seems that you didn't update the local fork before creating the pull request.

src/conv/k-g/k-g.cpp
```diff
@@ -131,58 +201,74 @@ int main
...
 				int n2 = kData.elements[*itr].nodes[1];
 				int n3 = kData.elements[*itr].nodes[2];
 
-				if (section > 0) {
-				    KSectionBeam beamSection = kData.sectionsBeam[section];
+				if (sectionid > 0) {
+				    //KSectionBeam ksectionBeam = kData.sectionsBeam[sectionid];
 
-				    if (beamSection.CST == 0) {
+				    if (ksectionBeam.CST == 0) {
```
AliHaider93, line 248: unrelated+ unnecessary

src/conv/k-g/k-g.cpp
```diff
@@ -131,58 +201,74 @@ int main
...
 				    }
-				    else if (beamSection.CST == 1 && beamSection.sectionType == "") {
+				    else if (ksectionBeam.CST == 1 && ksectionBeam.sectionType == "") {
```
AliHaider93, line 250: same

src/conv/k-g/k-g.cpp
```diff
@@ -131,58 +201,74 @@ int main
...
+					point1.outerDiameter = ksectionBeam.TS1;
+					point1.innerDiameter = ksectionBeam.TT1;
 
 					point2.coords[X] = kData.nodes[n2].x * factor;
 					point2.coords[Y] = kData.nodes[n2].y * factor;
 					point2.coords[Z] = kData.nodes[n2].z * factor;
-					point2.outerDiameter = beamSection.TS2;
-					point2.innerDiameter = beamSection.TT2;
+					point2.outerDiameter = ksectionBeam.TS2;
+					point2.innerDiameter = ksectionBeam.TT2;
```
AliHaider93, line 264: same

src/conv/k-g/k-g.cpp
```diff
@@ -131,58 +201,74 @@ int main
...
 
 					geometry.addPipePnt(point1);
 					geometry.addPipePnt(point2);
 				    }
-				    else if (beamSection.CST == 2) {
+				    else if (ksectionBeam.CST == 2) {
 				    }
-				    else if ((beamSection.sectionType != "") && (beamSection.sectionType != "SECTION_08") && (beamSection.sectionType != "SECTION_09")) {
+				    else if ((ksectionBeam.sectionType != "") && (ksectionBeam.sectionType != "SECTION_08") && (ksectionBeam.sectionType != "SECTION_09")) {
```
AliHaider93, line 271: same

src/conv/k-g/k_parser.cpp
```diff
@@ -478,13 +474,13 @@ bool parse_k
 			    if (command.size() == 1)
 				state = KState::Include;
 			    else
-				std::cout << "Unexpected command " << tokens[0] << " in k-file " << fileName << std::endl;
+				std::cout << "Unexpected command " << tokens[0] << " in k-file " << fileName << "\n";
```
AliHaider93, line 477: update the local fork before creating the pull request

PR #195, opened 2025-07-12
First Element_Discrete_Sphere implementation

BraghinAndrea:
This PR's purpose is to show the first Element_Discrete_Sphere implementation in k-g converter to collect feedbacks and reviews. It might be needed to wait for Seatbelt PR194 before merging

### DIFF
--- a/src/conv/k-g/CMakeLists.txt
+++ b/src/conv/k-g/CMakeLists.txt
@@ -8,6 +8,7 @@ set(k-g_soure_files
   pipe.cpp
   sketch.cpp
   extrude.cpp
+  sphere.cpp
   )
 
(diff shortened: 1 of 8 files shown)


brlcad (changes requested):
On closer inspection, this doesn't look right.  You're implementing sphere, but writing out a particle entity instead and a corrupt one at that (wrong magic for sphere, but right for particle).   Type you should be exporting is an rt_ell_internal instead of an rt_part_internal.

brlcad:
More concerning is whether this was tested as I would expect the type mismatch to cause a runtime assertion/bomb failure.  It's possible the ID_SPH is somehow compatible or not getting checked, or maybe not checked until time-of-use (e.g., trying to retrace the particle).

PR #197, opened 2025-08-27
Update k_parser.cpp

BraghinAndrea:
with this PR we can convert the Yaris model with no Errors. This is not final it's just to show the mentors what did I do and ask for feedback

### DIFF
--- a/src/conv/k-g/k_parser.cpp
+++ b/src/conv/k-g/k_parser.cpp
@@ -76,7 +76,7 @@ enum class KState {
     Section_Ale2d,
     Section_Beam,
     //Section_Beam_AISC,
-    //Section_Discrete,
+    Section_Discrete,
     //Section_Fpd,
     //Section_Point_Source,
     //Section_Point_source_Mixture,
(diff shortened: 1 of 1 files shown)


drossberg:
This pull request has the same issues as #203: https://github.com/BRL-CAD/brlcad/pull/203#issuecomment-3904799189

drossberg:
I close this PR, because @AliHaider93 checked and it can be discarded, all the changes are implemented in PR  #203 in a better way:
- Using SectionHasOption in #203 is the correct way #197 uses local hasoption.
- ICOMP option is handeled correctly in #203  numberOfCards += (nip + 7) / 8 while in #197 it's just numberOfCards += 1 which is not the case.
- PR #203 includes  lmc and nipp options while #197 only handles ELFORM and ICOMP.
- Section_solid keyword is fully handlded in #203 while #197 only handels the options written in the docs [image.png](/user_uploads/1549/WnaMjNGiHDWfk7kvH-QjnJGD/image.png).
- Element_Discrete_Sphere has it's full implementation in PR #203 nothing in #197.

PR #198, opened 2025-08-29
clean up for shpere

BraghinAndrea:
This is a clean version of the discrete sphere PR, looking forward for feedback

### DIFF
--- a/src/conv/k-g/sphere.cpp
+++ b/src/conv/k-g/sphere.cpp
@@ -1,0 +1,100 @@
+/*                 S P H E R E . C P P
+ * BRL-CAD
+ *
+ * Copyright (c) 2024-2025 United States Government as represented by
+ * the U.S. Army Research Laboratory.
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public License
+ * version 2.1 as published by the Free Software Foundation.
+ *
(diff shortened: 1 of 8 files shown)


src/conv/k-g/sphere.cpp
```diff
@@ -0,0 +1,100 @@
...
+Sphere::Sphere(void) : m_name() {}
+
+void Sphere::setName
+(
+    const char* value
+) {
+    if (value != nullptr)
+	m_name = value;
+    else
+	m_name = "";
```
brlcad, line 37: I didn't check if the other entities do this but there is at least conceptually a difference between a null name and an empty string name.  If this is exclusively used for object-creation on the wdb side of things, then write() should probably auto-generate some unique string name (e.g., "unnamed_001_" monotonically increasing) so that it's uniquely identifiable in the .g database.  Otherwise the object will get overwritten with other ""-named sphere objects.

src/conv/k-g/sphere.h
```diff
@@ -0,0 +1,61 @@
+/*                 P I P E . H
```
brlcad, line 1: This is wrong...

brlcad (approved):
Didn't see any major issues but there are some concerns and at least one minor issue that I left commented.  Nice to see the progress.  If you have a demo of this in action, would be good to see a screenshot in Zulip.

PR #203, opened 2025-10-05
Update k_parser.cpp

BraghinAndrea:
here is a the logic for handling section shell length and the error related to it in case of multiple sections under the same keyword

### DIFF
--- a/src/conv/k-g/k_parser.cpp
+++ b/src/conv/k-g/k_parser.cpp
@@ -76,7 +76,7 @@ enum class KState {
     Section_Ale2d,
     Section_Beam,
     //Section_Beam_AISC,
-    //Section_Discrete,
+    Section_Discrete,
     //Section_Fpd,
     //Section_Point_Source,
     //Section_Point_source_Mixture,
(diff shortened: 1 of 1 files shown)


drossberg:
There are several issues with this pull request:
1. Compilation error with gcc
```
src/conv/k-g/k_parser.cpp: In function ‘bool parse_k(const char*, KData&)’:
src/conv/k-g/k_parser.cpp:781:63: error: no matching function for call to ‘find(std::vector<std::__cxx11::basic_string<char> >::iterator, std::vector<std::__cxx11::basic_string<char> >::iterator, const char [6])’
  781 |                                     bool hasTitle = (std::find(command.begin(), command.end(), "TITLE") != command.end());
      |                                                      ~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
Including the header file, where std::find() is defined, fixes it: `include <algorithm>`
2. The variable `options` is declared multiple times: In lines 458, 784, and 814.
3. The `switch` statement starting at line 876 doesn't handle `Section_Discrete`. Changing back line 79 to `//Section_Discrete,`?

PR #204, opened 2025-10-16
added create_material

ajkomolafe:
takes material from get_shader and creates the material within the 3rd gen command

### DIFF
--- a/src/libgcv/plugins/rhino/rhino_read.cpp
+++ b/src/libgcv/plugins/rhino/rhino_read.cpp
@@ -459,19 +459,71 @@ write_geometry(rt_wdb &wdb, const std::string &name, const ON_Geometry *geometry
     return 1;
 }
 
+std::string create_material(const ON_Material *im, rt_wdb& wdb)
+{
+    ON_Material dm;
+	
+    const ON_Material *material = (im) ? im : &dm;
+	// dm->ToPhysicallyBased();
+	// look at later
+
+	// if the name already exists then add some prompt in the cmdline 
+	// to check if they want to overwrite, merge, or keep the old material
+
+	std::string name = clean_name(material->Name(), "plastic");
+
+	struct bu_attribute_value_set physicalProperties;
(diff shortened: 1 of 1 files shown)


ajkomolafe:
wrong repo

src/libgcv/plugins/rhino/rhino_read.cpp
```diff
@@ -459,20 +459,72 @@ write_geometry(rt_wdb &wdb, const std::string &name, const ON_Geometry *geometry
     return 1;
 }
 
+std::string create_material(const ON_Material *im, rt_wdb& wdb)
+{
+    ON_Material dm;
+	
+    const ON_Material *material = (im) ? im : &dm;
```
brlcad, line 466: a locally scoped ON_Material probably isn't helpful here.  if "im" is null, then there's nothing to create and we probably shouldn't have called create_material().

src/libgcv/plugins/rhino/rhino_read.cpp
```diff
@@ -459,20 +459,72 @@ write_geometry(rt_wdb &wdb, const std::string &name, const ON_Geometry *geometry
...
+	// dm->ToPhysicallyBased();
+	// look at later
+
+	// if the name already exists then add some prompt in the cmdline 
+	// to check if they want to overwrite, merge, or keep the old material
+
+	std::string name = clean_name(material->Name(), "plastic");
```
brlcad, line 473: "plastic" is a shader name (i.e., old system), which should really only be introduced/kept if we're creating old shader strings as the code was doing.  for newer mk_material() objects, a better default name might simply be "material" or something derived from the material object's parameters (e.g., if we know it's just a color, use the stored color name if there is one).

src/libgcv/plugins/rhino/rhino_read.cpp
ajkomolafe: 👍

PR #212, opened 2025-12-30
Fix stack buffer overflow in nastran-g PBAR/PSHELL name generation

### LINKED ISSUE #211: Stack Buffer Overflow in nastran-g.c when processing large PBAR/PSHELL PIDs
**Bug Type:** Security / Buffer Overflow / Crash

**Description:**
In `src/conv/nastran-g.c`, the PBAR and PSHELL conversion loops use `sprintf` to write group names into a fixed-size stack buffer:

```c
char name[NAMESIZE+1];  // 17 bytes
sprintf(name, "pbar_group.%d", pbp->pid);
sprintf(name, "pshell.%d", psh->pid);
```
If the PID has more than 6 digits (or 10 digits for 32-bit max int), the resulting string exceeds the buffer size, causing a stack buffer overflow. This can crash nastran-g and theoretically allow code execution.

### Proposed Fix:

Increase the buffer size to 32 bytes:
```c
char name[32];
```

Use snprintf instead of sprintf:
```c
snprintf(name, sizeof(name), "pbar_group.%d", pbp->pid);
snprintf(name, sizeof(name), "pshell.%d", psh->pid);
```

THE-Amrit-mahto-05:
### Description
PR fixes the stack buffer overflow in nastran-g.c when PBAR/PSHELL PIDs are large.
Uses snprintf and increases buffer sizes.

Fixes #211

### DIFF
--- a/src/conv/nastran-g.c
+++ b/src/conv/nastran-g.c
@@ -221,13 +221,13 @@ do_silly_nastran_shortcuts(void)
 
 		a = atof(prev_rec[field_no]);
 		b = atof(&curr_rec[field_no][i]);
-		sprintf(curr_rec[field_no], "%-#*E", FIELD_LENGTH-6, a+b);
+		snprintf(curr_rec[field_no], FIELD_LENGTH, "%-#*E", FIELD_LENGTH-6, a+b);
 	    } else {
 		int a, b;
 
 		a = atoi(prev_rec[field_no]);
 		b = atoi(&curr_rec[field_no][i]);
-		sprintf(curr_rec[field_no], "%d", a+b);
+		snprintf(curr_rec[field_no], FIELD_LENGTH, "%d", a+b);
 	    }
 	}
     }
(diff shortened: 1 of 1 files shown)


starseeker:
Replacing NAMESIZE with a hard coded number is iffy - the code doesn't obviously self-document why 32 is used there.

THE-Amrit-mahto-05:
hey @starseeker 
I used 32 because "pbar_group." + a large integer needs about 23 characters, so 32 was a safe size. But yeah, hardcoding it was definitely not a good idea. How would you prefer I fix this? Should I use struct bu_vls instead, or define a macro at the top?

starseeker:
@THE-Amrit-mahto-05  I think this may have gotten taken care of by https://github.com/BRL-CAD/brlcad/commit/17f62523250b3c8880417a4e546c18bc4ca3854e - can you confirm if that resolves the issue for you?

starseeker:
Sorry, actually that was https://github.com/BRL-CAD/brlcad/commit/ef61c69eb6d05941bb891ed4ad45b268bde1eaeb with the vls version.

starseeker:
Looks like the https://github.com/BRL-CAD/brlcad/commit/17f62523250b3c8880417a4e546c18bc4ca3854e might have been a duplicate of this PR - apologies if I used the wrong one as a basis for the vls fix.

PR #213, opened 2026-01-01
librt/nmg: guard pl_ray() debug plotting behind NMG_DEBUG_PLOTEM

AAYUSH-SPIDEY-SHARMA:
pl_ray() currently writes .plot3 files unconditionally when triggered from unresolved(), which can result in unexpected file output during NMG raytracing.

This change guards the pl_ray() call behind the existing NMG_DEBUG_PLOTEM flag (“make plots in debugged routines”), ensuring plot files are only generated when explicitly requested via NMG debug flags.

This preserves existing debugging functionality while avoiding default disk writes.


drossberg:
Hi @AAYUSH-SPIDEY-SHARMA, please, have a look at HACKING (in the root of the brlcad repository). See the "1) Indentation whitespace" section under "CODING STYLE & STANDARDS" there.

PR #220, opened 2026-02-26
gist: replace hardcoded render constants with named defaults

AAYUSH-SPIDEY-SHARMA:
Replace hardcoded render size ("1024") and ambient occlusion samples ("set ambSamples=8") literals in PerspectiveGatherer.cpp with file-local static constants.

No behavioral change. This prepares for future quality configuration work.


starseeker:
Could you rework this to apply to latest main?

Issue #161, opened 2024-09-19
rt_db_put_internal() does not work with inmem database


bylee20:
I have created empty database with db_create_inmem() and It seems that `rt_db_put_internal()` does not work with inmem database.
Here's my test code:

```
    const auto db = db_create_inmem();

    rt_ell_internal *sph = nullptr;
    BU_ALLOC(sph, rt_ell_internal);
    *sph = {
        .magic = RT_ELL_INTERNAL_MAGIC,
        .v = {0, 0, 0},
        .a = {1, 0, 0},
        .b = {0, 1, 0},
        .c = {0, 0, 1},
    };

    rt_db_internal in = {
        .idb_magic = RT_DB_INTERNAL_MAGIC,
        .idb_major_type = DB5_MAJORTYPE_BRLCAD,
        .idb_minor_type = ID_SPH,
        .idb_meth = &OBJ[ID_SPH],
        .idb_ptr = sph,
        .idb_avs = BU_AVS_INIT_ZERO,
    };

    const auto dir = db_diradd(db, "sph.o", RT_DIR_PHONY_ADDR, 0, RT_DIR_SOLID, &in.idb_minor_type);
    rt_db_put_internal(dir, db, &in, &rt_uniresource);
```

If you execute this code, you will see error message:

> rt_db_put_internal5(sph.o) db_realloc5() failed

I did short grepping and found that this error comes from:
https://github.com/BRL-CAD/brlcad/blob/02ba341f2ea0c58c896a9224ec8c6ed7b434f6db/src/librt/db5_io.c#L767-L773

which is result of failure of `db5_realloc()` call and again, this failure is originated from:
https://github.com/BRL-CAD/brlcad/blob/02ba341f2ea0c58c896a9224ec8c6ed7b434f6db/src/librt/db5_alloc.c#L135-L141

Here, `db_dirbuild()` call always fails for inmem database because of this check:
https://github.com/BRL-CAD/brlcad/blob/02ba341f2ea0c58c896a9224ec8c6ed7b434f6db/src/librt/db5_scan.c#L387-L389

Since dbip is for inmem database, the file pointer will be always null and db_dirbuild() never work.

I also found that wdb_put_internal() works fine with inmem database.

Is this by design or a bug? Or, did I do something wrong?

starseeker:
@brlcad do you have any insights about the design intent here?  Is rt_db_put_internal expected to work in this case?

bylee20:
[...]

PR #221, opened 2026-02-26
librt: fix in-memory database write failure

AWaleed-Ahmed:
This PR fixes a write-path failure in in-memory databases in BRL-CAD that occurred when rt_db_put_internal() triggered db5_realloc(), which then called db_dirbuild() and failed because it assumed a file-backed database and returned an error when dbi_fp == NULL. I updated db_dirbuild() to properly handle in-memory databases by delegating to db_dirbuild_inmem() or initializing a fresh empty state by setting dbi_eof = 0, adjusted db5_realloc() to safely allocate memory when no file pointer exists instead of attempting disk-based allocation, and corrected db_diradd() to preserve the RT_DIR_INMEM flag for memory-only databases. These changes resolve the initialization deadlock while maintaining full backward compatibility for file-backed and memory-mapped database #161


drossberg:
Hi @AWaleed-Ahmed, can you resolve the conflicts? Why is it necessary to set the RT_DIR_INMEM flag? There could be a good reason to not have set it, or? How have you tested your code?

AWaleed-Ahmed:
Hey @drossberg sorry for getting back with you a bit late, I am a bit occupied in a few prior commitments so I'll be resolving the conflicts this week when I get free. About the RT_DIR_INMEM flag, I used it because without it, the library defaults to file-offset logic even for in-memory databases which made db_write attempting to access a NULL file pointer. Also I traced the write path using static control-flow analysis to understand how the database transitions during initialization. I verified that db_dirbuild correctly updates dbi_eof from -1 to 0, indicating that the database has been built. I also confirmed that db5_realloc safely falls back to bu_malloc when dbi_fp is NULL.

PR #224, opened 2026-03-16
proc-db: add random CSG boolean geometry generator

rohits1450:
## Summary
Adds a procedural geometry generator `random_csg.c` to `src/proc-db` 
that creates random BRL-CAD primitives (spheres and boxes) combined 
using random boolean operations (union, subtract, intersect), writing 
the result to a `.g` database file.

## What it does
- Generates N random primitives at random positions and sizes
- Randomly assigns boolean operations (union/subtract/intersect)
- Writes a valid .g database readable by MGED
- Useful for testing geometry pipelines and BRL-CAD tooling

## Testing
Built and verified on Windows (Visual Studio 18 2026):
- Compiled successfully against libwdb
- Generated random_csg.g verified in MGED
- Rendered using `B random_csg.c` command in MGED
image

## Context
Developed as groundwork for the Python Bindings GSoC 2026 project,
aligned with the libwdb API approach for the binding layer.

### DIFF
--- a/src/proc-db/CMakeLists.txt
+++ b/src/proc-db/CMakeLists.txt
@@ -37,7 +37,7 @@ brlcad_addexec(wdb_example wdb_example.c libwdb NO_INSTALL) # NOTE: has manual p
 #
 # we still DO NOT INSTALL them because they do not have documentation
 # and impact our perceived quality and complexity.
-
+brlcad_addexec(random_csg random_csg.c "libwdb;${M_LIBRARY}" NO_INSTALL)
 brlcad_addexec(brickwall brickwall.c libwdb NO_INSTALL)
 brlcad_addexec(cchannel cchannel.c "libwdb;${M_LIBRARY}" NO_INSTALL)
 brlcad_addexec(clutter "clutter.c;common.c" "libwdb;${M_LIBRARY}" NO_INSTALL)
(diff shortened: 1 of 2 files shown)


brlcad:
@rohits1450 this is a nice demonstration and quick turn-around, but a few issues that give pause.  First the header, style, and footer don't match the other proc-db examples.  The sh/template.sh script can fix that for you automatically, or can copy the other programs.  Second, the choice of N appears to be hard-coded instead of user-input which makes this awkward to use for testing purposes.  I would have expected from the earlier discussion that the command would either take an argument for what N to use, or to even specify N-min to N-max for a range of # of primitives to randomly generate, and more interestingly to generate more than shapes than rpp and sph (ideally also user-specified) or with random rotations so the output geometry is more diverse.  Third, given a primary purpose would be for testing and demo purposes, do you think you could make it take another argument for the number of combs to make?  It would be visually interesting to plot them in a grid so each comb is modeled at the origin like you currently have it creating them, but then translated into a final position for a final scene comb (e.g., --min=1 --max=10 --count=100 would create a 10x10 grid of combs with 1 to 10 boolean ops) . 

Note:  while it's certainly fine if you used an LLM to help code this up, that would be something worth disclosing.

PR #227, opened 2026-03-21
fix: corrections for analyze methods for `ehy`, `hyp` and `rhc`

agkphysics:
Many years ago I contributed some volume/centroid/surface area calculations to this project as part of Google Code-in. However, some of my equations were wrong or inefficient and it's been on my mind for a while to fix them. So here are fixes for the `ehy`, `hyp` and `rhc` primitives. I've also added `rt_ehy_volume` which did not previously exist. I've compared these outputs with those from `check volume` and `check surf_area` and they are quite close.

Some of the calculations require approximation due to having integrals without a closed-form solution. In particular I have used Monte Carlo integration for the surface area integrals since they are double integrals, but I do not know the optimal number of steps to get good convergence.

### DIFF
--- a/src/libged/analyze/analyze.cpp
+++ b/src/libged/analyze/analyze.cpp
@@ -267,6 +267,10 @@ analyze_do_summary(struct ged *gedp, const struct rt_db_internal *ip)
 	    analyze_sketch(gedp, ip);
 	    break;
 
+	case ID_EHY:
+	    analyze_general(gedp, ip);
+	    break;
+
 	case ID_HYP:
 	    analyze_general(gedp, ip);
 	    break;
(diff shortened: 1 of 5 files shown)


starseeker:
@agkphysics  Heh - I actually went through a general pass updating vol and surf area methods as part of some facetization work, which looks like it conflicts with these changes.  Do you think you could check your updates against what is there in latest main and rebase your corrections if main is still wrong?

agkphysics:
@starseeker I suspect the Crofton method is probably more efficient than the Monte-Carlo method I implemented, but I will try and benchmark when I get time. I think at least `rt_hyp_volume` and `rt_rhc_centroid` still need to change.

PR #230, opened 2026-03-30
Fix ELEMENT_BEAM multi-card parsing in k-g converter

LorenzoSanti34:
## Summary
- Fix optionsCounter not incrementing after Card 1, causing option cards 
  (THICKNESS, SECTION, etc.) to be misrouted back to the Card 1 branch 
  with "Too short ELEMENT_BEAM" error
- Fix local variables (eid, element) resetting on every line, causing 
  option data to overwrite stored elements with empty data

## Test file
```
*KEYWORD
*TITLE
Test ELEMENT_BEAM_THICKNESS bug
$
*NODE
       1 0.000000000e+00 0.000000000e+00 0.000000000e+00
       2 1.000000000e+01 0.000000000e+00 0.000000000e+00
       3 0.000000000e+00 1.000000000e+00 0.000000000e+00
       4 2.000000000e+01 1.000000000e+01 0.000000000e+00
$
*MAT_ELASTIC
         1 7.850e-09 2.100e+05 3.000e-01
$
*SECTION_BEAM_TITLE
Beam Section 1
       1       1   1.0000000e+00   0.0000000e+00       1       0
   5.0000000e+00   5.0000000e+00   2.0000000e+00   2.0000000e+00   0.0000000e+00   0.0000000e+00
$
*PART
Beam Part
         1         1         1
$
*ELEMENT_BEAM_THICKNESS
       1       1       1       2       3       0       0       0       0       2
   6.0000000e+00   6.0000000e+00   3.0000000e+00   3.0000000e+00   0.0000000e+00
       2       1       2       4       3       0       0       0       0       2
   8.0000000e+00   8.0000000e+00   4.0000000e+00   4.0000000e+00   0.0000000e+00
$
*END
```

## Before fix
- Card 2 (options data) enters Card 1 branch → "Too short ELEMENT_BEAM" error
- THICKNESS parameters are silently lost

## After fix
- Option cards are correctly routed to the options branch
- Element data persists across cards via persistent elementId variable
- Both beam elements parse and render correctly

### DIFF
--- a/src/conv/k-g/k_parser.cpp
+++ b/src/conv/k-g/k_parser.cpp
@@ -1443,6 +1443,10 @@ bool parse_k
 
 			    pid = stoi(tokens[1]);
 			    data.parts[pid].elements.insert(eid);
+
+			    if (beamOptions.size() > 0)
+				++optionsCounter;
+
 			    break;
 			}
 			else if ((beamOptions.size() > 0)) {
@@ -1623,6 +1627,9 @@ bool parse_k
 				default:
 				    break;
 			    }
+
+			    if (optionsCounter >= beamOptions.size())
+				optionsCounter = 0;
 			}
 
 			break;
--- a/src/conv/k-g/k_parser.cpp
+++ b/src/conv/k-g/k_parser.cpp
@@ -446,6 +446,7 @@ bool parse_k
 	std::string              sectionTitle;
 	bool                     sectionHasOption = false;
 	int                      sectionId        = -1;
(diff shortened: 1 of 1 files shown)


AliHaider93:
The bug is confirmed, the option counter not incrementing after reading card 1 causes wrong interpretation of the option card resulting in fictitious Too short ELEMENT_BEAM error and the loss of the option card's information. The fix is also valid both regarding the increment of options counter and the persistence of the element Id. @drossberg

PR #231, opened 2026-03-31
Fix missing *3 stride in BoT normal array indexing in k-g converter

laithaldaeef:
The `add_normal()` function in `bot.cpp` indexes the flat-packed normals array
  without the required `* 3` stride. The sibling function `add_vertex()` 
  correctly uses `ret * 3`, and every other normal array access across BRL-CAD
  (e.g., `src/librt/primitives/bot/bot.c`, `src/conv/obj-g.c`) uses `index * 3`.

### DIFF
--- a/src/conv/k-g/bot.cpp
+++ b/src/conv/k-g/bot.cpp
@@ -66,9 +66,9 @@ static int add_normal
 
     // normal already there?
     for (; ret < static_cast<int>(bot.num_normals); ++ret) {
-	tmp[0] = bot.normals[ret];
-	tmp[1] = bot.normals[ret + 1];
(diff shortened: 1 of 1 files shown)


drossberg:
Hi @laithaldaeef, you found and fixed an actual bug. You patch looks good, except in line 70. It has to start with a tab, not 4 spaces, according our coding rules (see the HACKING file in brlcad's main directory). When you correct this, I'll accept your PR.

starseeker:
@drossberg should we go ahead and fix/merge this for the bug fix?

drossberg:
Right, you can accept this PR and fix the style issue afterwards.

PR #234, opened 2026-05-02
Fix typo in GeometryIO.tcl which causes model exports to fail

wishstudio:
Using the `File->Export` function in archer causes the following error (same behavior in both Windows and Linux):

```
wrong # args: should be "set varName ?newValue?"
    while executing
"set binpath [bu_dir bin] ]"
    (procedure "cadwidgets::geom_save" line 3)
    invoked from within
"cadwidgets::geom_save $mTargetCopy $mTarget $itk_component(ged)"
    (object "::.archer0" method "::ArcherCore::exportDb" body line 21)
    invoked from within
"::.archer0 exportDb"
    (in namespace inscope "::Archer" script line 1)
    invoked from within
"namespace inscope ::Archer {::.archer0 exportDb}"
    (menu invoke)
```

Turns out to be a typo in `GeometryIO.tcl`.

### DIFF
--- a/src/tclscripts/lib/GeometryIO.tcl
+++ b/src/tclscripts/lib/GeometryIO.tcl
@@ -176,7 +176,7 @@ proc geom_load {input_file gui_feedback} {
 # and may only want a subset.
 proc geom_save {input_file output_file db_component} {
 
-    set binpath [bu_dir bin] ]
+    set binpath [bu_dir bin]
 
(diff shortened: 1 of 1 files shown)


starseeker:
That's a genuine typo, but if I apply the fix and try an export it looks like another error comes up.  Can you see if you can address the subsequent error as well?

PR #235, opened 2026-05-02
Only call set_target_properties on perm-test when testing is enabled

wishstudio:
Otherwise CMake will complain if `BUILD_TESTING` is false:

```
CMake Error at src/tclscripts/CMakeLists.txt:142 (set_target_properties):
  set_target_properties Can not find target to add properties to: perm_test
```


starseeker:
Thanks - looks like this is OBE in current main code

Issue #228, opened 2026-03-23
[BUG] BRL-CAD dxf-g heap buffer overflow via unchecked POLYLINE face indices in DXF import


quart27219:
### Summary
A crafted DXF file can trigger a heap-buffer-overflow in the latest BRL-CAD/brlcad dxf-g importer. The bug is reachable through the normal command-line conversion path and is caused by attacker-controlled POLYLINE face indices being used directly to index the stored vertex table without bounds validation. This is a native C memory-corruption issue and can at least cause a reliable denial of service. 

### Details
The latest upstream BRL-CAD/brlcad still contains the DXF importer flaw associated with CVE-2022-0496. I rebuilt the current upstream tree from a fresh full clone and exercised the real dxf-g converter with a crafted DXF file through the normal import path.

Reachable source file:
- `brlcad/src/conv/dxf/dxf-g.c`

Entry path:
```
while ((code=readcodes()) > -900) {
    process_code[curr_state->state](code);
}
Relevant vulnerable fragment in src/conv/dxf/dxf-g.c:

case 71:
case 72:
case 73:
case 74:
    coord = (code % 70) - 1;
    face[coord] = abs(atoi(line));
    break;
case 0:
    get_layer();
    if (vertex_flag == POLY_VERTEX_FACE) {
        add_triangle(polyline_vert_indices[face[0]-1],
                     polyline_vert_indices[face[1]-1],
                     polyline_vert_indices[face[2]-1],
                     curr_layer);
```

The vulnerable condition is that face[] is populated directly from attacker-controlled DXF input, and then used as polyline_vert_indices[face[i]-1] without verifying that the face indices fall inside the actual vertex table.

This is a real parser-state-machine path, not a helper-only harness. The parser reaches the face-processing state normally, then dereferences out-of-range heap memory when the crafted face indices reference non-existent vertices.

### ASan LOG
ASan excerpt from the latest full-path rerun:
```
=================================================================
==67407==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x504000002c2c at pc 0x63103b0315e5 bp 0x7ffe242fd790 sp 0x7ffe242fd780
READ of size 4 at 0x504000002c2c thread T0
    #0 0x63103b0315e4 in process_entities_polyline_vertex_code /src/src/conv/dxf/dxf-g.c:764
    #1 0x63103b02181c in main /src/src/conv/dxf/dxf-g.c:3286
[...]

PR #245, opened 2026-07-14
dxf-g: Fix heap-buffer-overflow in polyline face generation

Meghna45-code:
This Pull Request fixes the heap-buffer-overflow crash described in BRL-CAD #228 .

The Issue:
1.When parsing a DXF file, BRL-CAD reads physical coordinates (corners) and stores them sequentially in a heap-allocated list called polyline_vert_indices.
2.When parsing a face connection instruction (using a vertex flagged with POLY_VERTEX_FACE), the translator reads the corner indices from the DXF file and saves them to the face array.
3.To build the mesh, BRL-CAD connects these corners by looking up the actual coordinates in the list: polyline_vert_indices[face[i] - 1]

The Vulnerability: The translator trusted the DXF file completely and did not check if the index specified (face[i]) was valid. If the DXF file contains an invalid ID (such as 0 which leads to index -1, or a very large index like 99999 when only a few vertices are defined), the program attempted to access memory outside the bounds of the polyline_vert_indices array, causing a heap-buffer-overflow and immediate crash.

I modified the process_entities_polyline_vertex_code function in dxf-g.c to validate the input indices before passing them to 
add_triangle:

1.Added bounds checks to ensure every parsed corner index (face[0], face[1], face[2], face[3]) is >= 1 and <= polyline_vert_indices_count.
2. If a face contains an out-of-bounds index, BRL-CAD now logs a warning and safely skips drawing that triangle instead of accessing invalid memory and crashing.

### DIFF
--- a/include/common.h
+++ b/include/common.h
@@ -101,6 +101,8 @@ double rint(double x);
 #ifndef THREADLOCAL
 #  if defined(__cplusplus) && __cplusplus >= 201103L
 #    define THREADLOCAL thread_local // C++11 or newer: thread_local is standard
+#  elif defined(_MSC_VER)
+#    define THREADLOCAL __declspec(thread) // MSVC C/C++ thread local storage
 #  elif !defined(__cplusplus) && defined(__STDC_VERSION__) && __STDC_VERSION__ >= 202311L
 #    define THREADLOCAL thread_local // C23: thread_local is standard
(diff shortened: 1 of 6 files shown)


starseeker:
Can you separate just the dxf-g fix into its own PR?  THREADLOCAL and build target names are separate discussions, and I think we already got the BOT fix in a different PR?

PR #249, opened 2026-08-17
workarround for make install DESTDIR var wrong behavior

### LINKED ISSUE #248: Cannot compile 7.44.0 in Linux
I'm unable to compile v7.44.0 on Ubuntu 24.04 64bit. I always get many errors like these:
...
dpkg-shlibdeps: error: cannot find library libz_brl.so.1 needed by debian/brlcad/usr/brlcad/bin/adrt_slave (ELF format: 'elf64-x86-64' abi: 'ELF:64:l:amd64:0'; RPATH: '/usr/brlcad/lib:/usr/brlcad/bin/../lib')
dpkg-shlibdeps: error: cannot find library libpng_brl16.so.16 needed by debian/brlcad/usr/brlcad/libexec/dm/libdm-X.so (ELF format: 'elf64-x86-64' abi: 'ELF:64:l:amd64:0'; RPATH: '/usr/brlcad/lib:/usr/brlcad/libexec/dm/../../lib')
dpkg-shlibdeps: warning: cannot find library libOpenNURBS.so needed by debian/brlcad/usr/brlcad/bin/naca456 (ELF format: 'elf64-x86-64' abi: 'ELF:64:l:amd64:0'; RPATH: '/usr/brlcad/lib:/usr/brlcad/bin/../lib')
dpkg-shlibdeps: error: cannot find library libpng_brl16.so.16 needed by debian/brlcad/usr/brlcad/libexec/dm/libdm-ogl.so (ELF format: 'elf64-x86-64' abi: 'ELF:64:l:amd64:0'; RPATH: '/usr/brlcad/lib:/usr/brlcad/libexec/dm/../../lib')
dpkg-shlibdeps: error: cannot continue due to the errors listed above
Note: libraries are not searched in other binary packages that do not have any shlibs or symbols file.
To help dpkg-shlibdeps find private libraries, you might need to use -l.
...

Adding -l"path" to dh_shlibdeps does not help

Just as reference, I've successfully compiled BRL-CAD 7.42.0 in the same system

jordisayol:


starseeker:
@jordisayol Thanks!  One question about the brlcad-v4 vs. brlcad-v5 bit - since they're both .g, do we need to call out v4 separately?  Is Debian actually spotting v4 vs. v5 files as different types?

jordisayol:
@starseeker 

One question about the brlcad-v4 vs. brlcad-v5 bit - since they're both .g, do we need to call out v4 separately?

Well, if I don't misremembering, Morrison told me that there are two brlcat data base, v4 and v5, both with the extension .g, For that we created two separated types with magic, because they have some fixed patrons on the beginning of the files

 Is Debian actually spotting v4 vs. v5 files as different types?

Yes, debian clearly distinguish between one an the other, and assigns different icons for each type, and even you can assign run different programs for each type, if you want

Cliff, if you think that there are not any geometry database v4 out there, or you want to rethink all this, please let me know

jordisayol:
@starseeker but they are not different types, they are same type with different versions, and this is shown on properties:
`Type: BRL-CAD Geometry Database v4 (application/brlcad-v4)`
and
`Type: BRL-CAD Geometry Database v5 (application/brlcad-v5)`

Another thing. Using magic and not only the .g extension to detect the file type, helps to be first against other programs that try to use the .g extension as their own type. I sow this in the past, don't remember details

starseeker:
Cool - there are definitely v4 databases out there, so that's fine.  It's neat that it can do that - I wasn't sure what their system was capable of.  I'll get this merged and migrate it over to the rel-7-44 branch so our 7.44.2 patch will be able to produce a deb as well - thanks!

starseeker:
@jordisayol merged, and added to rel-7-44 branch as well.  Let me know if there are any issues in rel-7-44 specifically and we'll try to get them ironed out before patch time.

jordisayol:
The rel-7-44 branch compiles to deb perfectly

PR #250, opened 2026-08-20
Update deb installation scripts

jordisayol:


starseeker:
Thanks!

jordisayol:
@starseeker thanks to you. I've done a small mistake on this PR. It is not necessary to create a debian/prerm script because de behavior is exactly the same in both cases.
I can do a new PR. Sorry for that

jordisayol:
@starseeker after read many about this thing, I'll keep as is because the prerm is guarantied to have the dependent packages present on system, but postrm script has not guaranteed this.

I just will add crudini in Pre-Depends field.

PR #251, opened 2026-08-21
Update dependency for crudini to Pre-Depends

jordisayol:
Changed dependency from Depends to Pre-Depends for crudini.


jordisayol:
@starseeker Thank you

